Issue #6, opened 2019-01-24
[QUESTION] Are BackendGroups extensible?


kierdavis:
From my understanding of how j5 is currently structured (please correct me if I'm wrong), it looks like a `BackendGroup` must have knowledge of every type of board that could provide. It also doesn't look like `HardwareBackendGroup` (and presumably other types of `BackendGroup` too, once they are defined) is competition-specific, so a `HardwareBackendGroup` therefore needs to have every board used in every competition hardcoded into it. Additionally, I'm not certain but it sounds like this structure would restrict the ability for end users to define their own boards using the j5 API.

If this is indeed true, my proposal would be for `BackendGroup`s to be initially empty, and to have each `Backend` register itself into its corresponding `BackendGroup` at the time of class definition.

trickeydan:
I like this idea and can definitely see the benefits that it would offer. When I get a chance I will have a think to consider how this could be done with the current project structure, or what changes would need to be made.

trickeydan:
I have solved this, will be submitting a PR when it's done.

PR #10, opened 2019-01-31
Auto register backends in BackendGroups

trickeydan:
This PR does make a fair amount of changes to the structure to the project.

- `Backend`s are now auto-registered with a `BackendGroup`
- A `Backend` now requires a `Board` and `Group` property
- It's more difficult to make errors
- A `BackendGroup` is now an instance of `BackendGroup`, not a subclass.

These changes fix the issues raised in #6


j5/backends/base.py
```diff
@@ -1,10 +1,44 @@
...
+
+    Responsible for registering the backend with a BackendGroup.
+
+    """
+
+    def __new__(mcs, name, bases, namespace, **kwargs):
+        """Create a new class object."""
+        cls = super().__new__(mcs, name, bases, namespace, **kwargs)  # type: ignore
+
+        if cls.__name__ == "Backend":
```
RealOrangeOne: You should be able to do `if isinstance(cls, Backend)` here, rather than relying on the class name directly

trickeydan: That wouldn't work? `Backend` is defined after `BackendMeta`, and is thus not in scope?

trickeydan: Although I could just be being tired.

PR #25, opened 2019-02-03
Fix board indices

### LINKED ISSUE #13: [FEATURE] Fix integer indices on boards being non-deterministic
- Remove integer indices
- Sort by serial number before looking.

pringon:
Fixes #13

### DIFF
--- a/j5/boards/base.py
+++ b/j5/boards/base.py
@@ -83,7 +83,7 @@ class BoardGroup:
 
     def update_boards(self) -> None:
         """Update the boards in this group to see if new boards have been added."""
-        self.boards = {}  # type: Dict[str, Board]
+        self.boards: Dict[str, Board] = {}
         for board in self.board_class.discover(self._backend):
             self.boards.update({board.serial: board})
 
--- a/j5/boards/base.py
+++ b/j5/boards/base.py
@@ -85,7 +85,7 @@ class BoardGroup:
         """Update the boards in this group to see if new boards have been added."""
         self.boards = {}  # type: Dict[str, Board]
         for board in self.board_class.discover(self._backend):
-            self.boards.update(board.serial: board)
+            self.boards.update({board.serial: board})
 
     def singular(self) -> Board:
         """If there is only a single board in the group, return that board."""
--- a/j5/boards/base.py
+++ b/j5/boards/base.py
@@ -118,11 +118,11 @@ class BoardGroup:
         return board
 
     def __getitem__(self, serial: str):
-        """Get the board from an index."""
+        """Get the board from serial."""
         if len(self.boards) <= 0:
             raise IndexError("Could not find any boards.")
         if type(serial) != str:
-            raise TypeError("Index must be a string")
+            raise TypeError("Serial must be a string")
         if serial not in self.boards:
             raise KeyError(f"Could not find a board with the serial {serial}")
         return self.boards[serial]
--- a/j5/boards/base.py
+++ b/j5/boards/base.py
@@ -85,12 +85,12 @@ class BoardGroup:
         """Update the boards in this group to see if new boards have been added."""
         self.boards = {}  # type: Dict[str, Board]
         for board in self.board_class.discover(self._backend):
-            self.boards[board.serial] = board
+            self.boards.update(board.serial: board)
 
     def singular(self) -> Board:
         """If there is only a single board in the group, return that board."""
         if len(self) == 1:
-            return self.boards[list(self.boards.keys())[0]]
+            return list(self.boards.values())[0]
         raise Exception("There is more than one or zero boards connected.")
 
     def make_safe(self):
@@ -113,7 +113,7 @@ class BoardGroup:
         """Get the next item in the iteration."""
         if self._iterator_counter >= len(self.boards):
             raise StopIteration
-        board = self.boards[list(self.boards.keys())[self._iterator_counter]]
+        board = list(self.boards.values())[self._iterator_counter]
         self._iterator_counter += 1
         return board
 
@@ -122,7 +122,7 @@ class BoardGroup:
         if len(self.boards) <= 0:
             raise IndexError("Could not find any boards.")
         if type(serial) != str:
-            raise KeyError("Index must be a string")
+            raise TypeError("Index must be a string")
         if serial not in self.boards:
             raise KeyError(f"Could not find a board with the serial {serial}")
         return self.boards[serial]
--- a/j5/boards/base.py
+++ b/j5/boards/base.py
@@ -1,11 +1,9 @@
 """The base classes for boards and group of boards."""
 
 from abc import ABCMeta, abstractmethod
-from typing import Iterator, Union
+from typing import Iterator
 
 from j5.backends import Backend
-
-BoardIndex = Union[int, str]
 
 
 class Board(metaclass=ABCMeta):
@@ -95,5 +93,3 @@ class BoardGroup:
         if serial not in self.boards:
             raise KeyError(f"Could not find a board with the serial {serial}")
         return self.boards[serial]
-
-
--- a/j5/boards/base.py
+++ b/j5/boards/base.py
@@ -64,7 +64,7 @@ class BoardGroup:
     def singular(self) -> Board:
         """If there is only a single board in the group, return that board."""
         if len(self) == 1:
-            return self.boards[0]
+            return self.boards[list(self.boards.keys())[0]]
         raise Exception("There is more than one or zero boards connected.")
 
     def __len__(self):
@@ -82,7 +82,7 @@ class BoardGroup:
         """Get the next item in the iteration."""
         if self._iterator_counter >= len(self.boards):
             raise StopIteration
-        board = self.boards[self.boards.keys()[self._iterator_counter]]
+        board = self.boards[list(self.boards.keys())[self._iterator_counter]]
         self._iterator_counter += 1
         return board
 
@@ -90,6 +90,8 @@ class BoardGroup:
         """Get the board from an index."""
         if len(self.boards) <= 0:
             raise IndexError("Could not find any boards.")
+        if type(serial) != str:
+            raise KeyError("Index must be a string")
         if serial not in self.boards:
             raise KeyError(f"Could not find a board with the serial {serial}")
         return self.boards[serial]
--- a/j5/boards/base.py
+++ b/j5/boards/base.py
@@ -1,15 +1,12 @@
 """The base classes for boards and group of boards."""
 
 from abc import ABCMeta, abstractmethod
-from typing import TYPE_CHECKING, Dict, Iterator, List, Type, Union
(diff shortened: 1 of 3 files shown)


trickeydan:
I'm going to resolve the conflicts so that I can review.

j5/boards/base.py
```diff
@@ -80,18 +77,20 @@ def __init__(self, board: Board, backend: Backend):
...
-        self.boards = self.board_class.discover(self._backend)
+        self.boards = {}  # type: Dict[str, Board]
+        for board in self.board_class.discover(self._backend):
+            self.boards[board.serial] = board
 
     def singular(self) -> Board:
         """If there is only a single board in the group, return that board."""
         if len(self) == 1:
-            return self.boards[0]
+            return self.boards[list(self.boards.keys())[0]]
```
trickeydan: Please replace with
```suggestion
            return list(self.boards.values())[0]
```

j5/boards/base.py
```diff
@@ -80,18 +77,20 @@ def __init__(self, board: Board, backend: Backend):
...
         self._iterator_counter: int = 0
-        self.boards: List[Board] = []
+        self.boards: Dict[str, Board] = {}
 
         self.update_boards()
 
     def update_boards(self) -> None:
         """Update the boards in this group to see if new boards have been added."""
-        self.boards = self.board_class.discover(self._backend)
+        self.boards = {}  # type: Dict[str, Board]
```
trickeydan: Can we use post Python 3.5 styles typing hints here please. I'm not entirely sure if we even need to redefine the type here as it is already defined in the constructor

j5/boards/base.py
```diff
@@ -80,18 +77,20 @@ def __init__(self, board: Board, backend: Backend):
...
+        for board in self.board_class.discover(self._backend):
+            self.boards[board.serial] = board
```
trickeydan: `self.boards.update` would be more pythonic here imo.

```suggestion
            self.boards.update(board.serial: board)
```

j5/boards/base.py
```diff
@@ -114,21 +113,16 @@ def __next__(self) -> Board:
         """Get the next item in the iteration."""
         if self._iterator_counter >= len(self.boards):
             raise StopIteration
-        board = self.boards[self._iterator_counter]
+        board = self.boards[list(self.boards.keys())[self._iterator_counter]]
```
trickeydan: Please can we use `self.boards.values()`

```suggestion
        board = list(self.boards.values())[self._iterator_counter]
```

j5/boards/base.py
```diff
@@ -114,21 +113,16 @@ def __next__(self) -> Board:
...
-            return self.boards[cast(int, index)]
-        elif type(index) == str:
-            for b in self.boards:
-                if b.serial == index:
-                    return b
-            raise KeyError(f"Could not find a board with the serial {index}")
-        else:
-            raise IndexError(f"Cannot index boards with type {str(type(index))}")
+        if type(serial) != str:
+            raise KeyError("Index must be a string")
```
trickeydan: This should be a `TypeError`
```suggestion
            raise TypeError("Index must be a string")
```

trickeydan: We have changed the name `index` to `serial` everywhere else.
```suggestion
            raise KeyError("Serial must be a string")
```

tests/boards/test_base.py
```diff
@@ -19,7 +19,7 @@ def name(self) -> str:
     @property
     def serial(self) -> str:
         """Get the serial number of this board."""
-        return "SERIAL"
+        return str(id(self))
```
trickeydan: This is better unit testing, but is also non-deterministic :(

tests/boards/test_base.py
```diff
@@ -152,15 +142,15 @@ def test_board_group_boards():
     board_group = BoardGroup(MockBoard, OneBoardMockBackend())
 
     assert len(board_group.boards) == 1
-    assert type(board_group.boards[0]) == MockBoard
+    assert type(list(board_group.boards.values())[0]) == MockBoard  # noqa: E501
```
trickeydan: Split these over a line rather than using `# noqa: E501`

tests/boards/test_base.py
```diff
@@ -152,15 +142,15 @@ def test_board_group_boards():
...
 
 
 def test_board_group_boards_multiple():
     """Test that the boards property works on multiple boards."""
     board_group = BoardGroup(MockBoard, TwoBoardsMockBackend())
 
     assert len(board_group.boards) == 2
-    assert type(board_group.boards[0]) == MockBoard
+    assert type(list(board_group.boards.values())[0]) == MockBoard  # noqa: E501
```
trickeydan: Split these over a line rather than using `# noqa: E501`

tests/boards/test_base.py
```diff
@@ -169,46 +159,35 @@ def test_board_group_boards_zero():
...
     board_group = BoardGroup(MockBoard, OneBoardMockBackend())
 
+    with pytest.raises(KeyError):
+        board_group[0]
+
     with pytest.raises(KeyError):
         board_group[""]
 
-    with pytest.raises(IndexError):
+    with pytest.raises(KeyError):
```
trickeydan: If you change the above comment from `KeyError` to `TypeError`, I think you need to change it here too.
```suggestion
    with pytest.raises(TypeError):
```

j5/boards/base.py
```diff
@@ -114,21 +113,16 @@ def __next__(self) -> Board:
...
-        if type(index) == int:
-            return self.boards[cast(int, index)]
-        elif type(index) == str:
-            for b in self.boards:
-                if b.serial == index:
-                    return b
-            raise KeyError(f"Could not find a board with the serial {index}")
-        else:
-            raise IndexError(f"Cannot index boards with type {str(type(index))}")
+        if type(serial) != str:
```
Skybound1: Do we need to validate that serial is a `str`?

j5/boards/base.py
```diff
@@ -114,21 +113,16 @@ def __next__(self) -> Board:
...
             raise StopIteration
-        board = self.boards[self._iterator_counter]
+        board = self.boards[list(self.boards.keys())[self._iterator_counter]]
         self._iterator_counter += 1
         return board
 
-    def __getitem__(self, index: BoardIndex) -> Board:
+    def __getitem__(self, serial: str):
         """Get the board from an index."""
         if len(self.boards) <= 0:
```
Skybound1: Is it worth removing this check as well?

pringon: It does seem a little redundant since the if statement on line 126 will cover that as well. Are there any situations where it would be useful to know that no boards are connected rather than the board not being found?

j5/boards/base.py
```diff
@@ -114,21 +113,16 @@ def __next__(self) -> Board:
...
-        elif type(index) == str:
-            for b in self.boards:
-                if b.serial == index:
-                    return b
-            raise KeyError(f"Could not find a board with the serial {index}")
-        else:
-            raise IndexError(f"Cannot index boards with type {str(type(index))}")
+        if type(serial) != str:
+            raise KeyError("Index must be a string")
+        if serial not in self.boards:
```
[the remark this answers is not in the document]
pringon: The KeyError that python throws doesn't seem informative enough.

For the code:
`random_dict={'a': 2, 'b': 3}
print(random_dict['c'])
`
It throws: 

> KeyError: 'c'

If you guys feel like that would be enough then I'd be happy to remove that check.

j5/boards/base.py
pringon: We could declare it at the beginning of the class before the constructor so that it would be more visible in the future. Not sure if this would be the 'python way' of doing things tho'.
Opinion?

j5/boards/base.py
pringon: Yes, otherwise the user may use a integer(for example). And instead of receiving a TypeError letting him know that he can only use strings, he would see the KeyError with the message "Could not find a board with the serial {serial}"

pringon:
Did I miss any suggested changes?

j5/boards/base.py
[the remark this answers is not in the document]
Skybound1: If that is the case, then I would be tempted to replace it with a try except block

```python
try:
    return self.boards[serial]
except KeyError:
    raise KeyError(f"Could not find a board with the serial {serial}") from None
```

The `from None` is to not have the "during handling of above exception"

j5/boards/base.py
Skybound1: @trickeydan can you think of any? My thoughts are that the object already has `__len__` in place, so we shouldn't _need_ to do that here. But if people decide they'd rather have it here as well, I'm fine with that too. And typically, `__getitem__` for python doesn't report the object is empty.

j5/boards/base.py
Skybound1: I would think the user would realise that they need to provide a serial with that error message as I think serials are typically strings. 

I do get your point with having the different exception though, is it worth moving this into the try-except clause? Where when we get the KeyError we then check if its the correct type and raise TypeError there?

j5/boards/base.py
trickeydan: Try `self.boards: Dict[str,Board] = {}`

j5/boards/base.py
trickeydan: Remember that we are trying to make the error messages as obvious as possible. Users of this API may have little knowledge of python.

j5/boards/base.py
trickeydan: So there are cases when the `BoardGroup` will be empty. However, as we are removing the ability to index by integer here, I would be fine with the removal of this check.

trickeydan:
> Did I miss any suggested changes?

There's a couple of unresolved conversations that I've now commented on. Also remember to merge master into this branch.

j5/boards/base.py
pringon: Would it really be necessary to move it into the except clause? I feel that it is more readable to be left out of it than perform the if statement inside the except. 
I can do it if you guys feel that it would be the better option.

j5/boards/base.py
pringon: I'm still not sure if type hinting is even necessary there since we already declared self.boards in the constructor which is just a few lines of code above the update_boards method. I will include the change now but I will like your feedback on this matter.

j5/boards/base.py
trickeydan: Ok, so I'm just going to make a decision on this one. Both solutions work, but @Skybound1's suggestions are more pythonic, so let's go with that.

pringon:
Is there anything else I need to do on this one?

j5/boards/base.py
kierdavis: @trickeydan The syntax that this line has been changed to seems more convoluted than the original - what would you say makes it more Pythonic?

trickeydan:
This is too complex to rebase and merge unfortunately.

PR #26, opened 2019-02-03
Revert "Rename supported_components to components"

kierdavis:
This reverts commit aa4f448a8096069c3115065c11627e61acd76533.

`components` would imply that the function returns every component
instance on the board, instead of the classes of component that
are supported by the board. `supported_components` better describes
what this function does.

This also updates the docstrings associated with the function to be clearer.


kierdavis:
Thanks

PR #27, opened 2019-02-03
Add ability to make robot safe.

### LINKED ISSUE #16: [FEATURE] Make Safe.
Things should be nice and safe.

trickeydan:
Fixes #16

### DIFF
--- a/j5/base_robot.py
+++ b/j5/base_robot.py
@@ -1,7 +1,5 @@
 """A base class for robots."""
 
-import atexit
-
 from j5.boards import Board
 
 
@@ -10,11 +8,4 @@ class BaseRobot:
 
     def make_safe(self):
         """Make this robot safe."""
-        BaseRobot.make_all_safe()
-
-    @staticmethod
-    @atexit.register
-    def make_all_safe():
-        """Make all boards safe."""
-        for board in Board.BOARDS:
-            board.make_safe()
+        Board.make_all_safe()
--- a/j5/base_robot.py
+++ b/j5/base_robot.py
@@ -1,0 +1,19 @@
+"""A base class for robots."""
+
+import atexit
+
+from j5.boards import Board
+
+
+class BaseRobot:
+
+    def make_self(self):
+        """Make this robot safe."""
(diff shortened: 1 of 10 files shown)


j5/base_robot.py
```diff
@@ -0,0 +1,20 @@
...
+
+import atexit
+
+from j5.boards import Board
+
+
+class BaseRobot:
+    """A base robot."""
+
+    def make_self(self):
```
sedders123:
```suggestion
    def make_safe(self):
```

j5/boards/base.py
```diff
@@ -23,6 +25,14 @@ def __repr__(self) -> str:
         """A representation of this board."""
         return f"<{self.__class__.__name__} serial={self.serial}>"
 
+    def setup(self):
```
sedders123: Requires caller to remember to call this. Is there a way to make this mandatory? (May not be just a thought)

j5/base_robot.py
```diff
@@ -0,0 +1,20 @@
...
+        """Make this robot safe."""
+        BaseRobot.make_all_safe()
+
+    @staticmethod
+    @atexit.register
+    def make_all_safe():
```
kierdavis: Seeing as this method is static, would it make more sense to move this method to the `Board` class?

j5/boards/base.py
```diff
@@ -11,6 +11,8 @@
 class Board(metaclass=ABCMeta):
     """A collection of hardware that has an implementation."""
 
+    BOARDS: List['Board'] = []  # All the boards.
```
kierdavis: Please can we have a comment to explain what the purpose of this list is?

e.g. "This list contains every instance of `Board` that has had its `setup` method called, and is used by `BaseRobot.make_all_safe` when Python exits to ensure every board is made safe."

j5/boards/base.py
trickeydan: Not easily, documented in #28 for future fixing as discussed.

j5/base_robot.py
trickeydan: I agree.

PR #40, opened 2019-02-04
Add Piezo component

HU90m:


j5/components/buzzer.py
```diff
@@ -0,0 +1,30 @@
...
+
+    def __init__(self, identifier: int, board: Board, backend: BuzzerInterface) -> None:
+        self._board = board
+        self._backend = backend
+        self._identifier = identifier
+
+    @staticmethod
+    def interface_class():
+        """Get the interface class that is required to use this component."""
+        return BuzzerInterface
```
kierdavis: This should probably have a `buzz` method (that calls `buzz` on the backend)

j5/components/buzzer.py
```diff
@@ -0,0 +1,30 @@
...
+from j5.boards import Board
+from j5.components import Component
+
+
+class BuzzerInterface(metaclass=ABCMeta):
+    """An interface containing the methods required to control an buzzer."""
+
+    @abstractmethod
+    def buzz(self, board: Board, identifier: int, length: int,
+             frequency: int = None, note: str = None) -> None:
```
trickeydan: Should we create a `Pitch` Type so that only one is needed. 
In the `Pitch` type, we should `Union` an `Enum` of note letters, and a int for the frequency.

If we change the name to `Piezo`, then a frequency is required.

j5/components/buzzer.py
```diff
@@ -0,0 +1,30 @@
...
+        """Set the state of an buzzer."""
```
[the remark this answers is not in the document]
kierdavis: Presumably `length` is the duration - can we rename that to `duration`, and perhaps make it a `datetime.timedelta`?

j5/components/buzzer.py
```diff
@@ -0,0 +1,30 @@
...
+        raise NotImplementedError  # pragma: no cover
+
+
+class Buzzer(Component):
```
trickeydan: Let's change this to `Piezo`

kierdavis: (For the record, the reasoning is that buzzers and piezos are slightly different concepts. A buzzer will produce a continuous tone in response to a continuous voltage being applied, while a piezo requires a PWM signal. Hence the pitch of a piezo can be varied by varying the PWM frequency, while the pitch of a buzzer is fixed.)

trickeydan:
Fixes #60

j5/components/piezo.py
```diff
@@ -0,0 +1,60 @@
...
+from typing import TypeVar
+
+from j5.boards import Board
+from j5.components import Component
+
+
+class Note(Enum):
+    """An enumerator of frequencies in Hz."""
+
+    C6 = 1046.50
```
trickeydan: In my opinion, these values should probably be integers

j5/components/piezo.py
```diff
@@ -0,0 +1,60 @@
...
+from datetime import timedelta
+from enum import Enum
+from typing import TypeVar
+
+from j5.boards import Board
+from j5.components import Component
+
+
+class Note(Enum):
+    """An enumerator of frequencies in Hz."""
```
trickeydan:
```suggestion
    """An enumeration of notes from scientific pitch notation and their related frequencies in Hz."""
```

j5/components/piezo.py
```diff
@@ -0,0 +1,60 @@
...
+    C7 = 2093.00
+    D7 = 2349.32
+    E7 = 2637.02
+    F7 = 2793.83
+    G7 = 3135.96
+    A7 = 3520.00
+    B7 = 3951.07
+
+
+Pitch = TypeVar('Pitch', int, Note)
```
trickeydan:
```suggestion
Pitch = Union[int, Note]
```

j5/components/piezo.py
```diff
@@ -0,0 +1,60 @@
+"""Classes for Piezo support."""
+
+from abc import ABCMeta, abstractmethod
+from datetime import timedelta
+from enum import Enum
+from typing import TypeVar
```
trickeydan:
```suggestion
from typing import Union
```
See below.

j5/components/piezo.py
HU90m: I couldn't find the use case for 'TypeVar' vs 'Union'. What are the differences?

sedders123: `TypeVar`s are generally used with [Generics](https://mypy.readthedocs.io/en/latest/generics.html#defining-generic-classes) whereas `Union`s are used to state that it can be one of the specified types.

HU90m: Thank you 😊

kierdavis:
I would prefer it if the frequencies were rounded to integers instead of truncated/floored, but I don't think that's a blocker - it's highly unlikely the difference will be noticeable in practice.

PR #56, opened 2019-02-06
Ensure board's are setup by overriding __new__

### LINKED ISSUE #28: No check for setup method call
The `board.setup()` method can easily be not called by the vendor, leading to a lack of safety for that board.

Possible fixes:
- Use `__new__`
- In the constructor (causes the same issue as super())
- Metaclass

sedders123:
This PR overrides the `__new__` method of `Board` to ensure that all instances have `.setup()` called on them.

I have also added a test to ensure that a `Board` can be correctly instantiated using all methods of variable declaration (positional, keyword arg and default). Happy for this to be removed if this is felt to be excessive.

There is also now the question of should `.setup()` be made "private" to discourage consumers from calling it?

All feedback welcome.

Closes #28

### DIFF
--- a/j5/boards/base.py
+++ b/j5/boards/base.py
@@ -27,14 +27,14 @@ class Board(metaclass=ABCMeta):
     def __new__(cls, *args, **kwargs):
         """Ensure any instantiated board is correctly setup."""
         instance = super().__new__(cls)
-        instance.setup()
+        instance._setup()
         return instance
 
     def __repr__(self) -> str:
         """A representation of this board."""
         return f"<{self.__class__.__name__} serial={self.serial}>"
 
-    def setup(self):
+    def _setup(self):
         """
         Setup the board.
 
--- a/j5/boards/base.py
+++ b/j5/boards/base.py
@@ -25,27 +25,14 @@ class Board(metaclass=ABCMeta):
         return f"{self.name} - {self.serial}"
 
     def __new__(cls, *args, **kwargs):
-        """Ensure any instantiated board is correctly setup."""
+        """Ensure any instantiated board is added to the boards list."""
         instance = super().__new__(cls)
-        instance._board_setup = False
-        instance._setup()
+        Board.BOARDS.append(instance)
         return instance
 
     def __repr__(self) -> str:
         """A representation of this board."""
         return f"<{self.__class__.__name__} serial={self.serial}>"
-
-    def _setup(self):
-        """
-        Setup the board.
-
-        Adds the implementation to BOARDS.
-        """
-        if not self._board_setup:
-            Board.BOARDS.append(self)
-            self._board_setup = True
-        else:
-            raise Exception(f"Setup has already been called for board {self.serial}")
 
     @property
     @abstractmethod
--- a/j5/boards/base.py
+++ b/j5/boards/base.py
@@ -24,6 +24,12 @@ class Board(metaclass=ABCMeta):
         """A string representation of this board."""
         return f"{self.name} - {self.serial}"
 
+    def __new__(cls, *args, **kwargs):
+        """Ensure any instantiated board is correctly setup."""
+        instance = super().__new__(cls)
+        instance.setup()
+        return instance
+
     def __repr__(self) -> str:
         """A representation of this board."""
         return f"<{self.__class__.__name__} serial={self.serial}>"
--- a/j5/boards/base.py
+++ b/j5/boards/base.py
@@ -27,6 +27,7 @@ class Board(metaclass=ABCMeta):
     def __new__(cls, *args, **kwargs):
         """Ensure any instantiated board is correctly setup."""
         instance = super().__new__(cls)
+        instance._board_setup = False
         instance._setup()
         return instance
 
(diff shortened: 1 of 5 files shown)


sedders123:
Any thoughts on making `.setup()` private (prefixing with an underscore)? My gut feeling is that it's a good idea, though not sure if you have any other plans for it.

If we do leave it "public", and probably even if we don't, it would be a good idea to guard against it being called multiple times, though I'm not sure what the correct behaviour here would be? Should it raise an exception, silently just do nothing, or maybe something else?

trickeydan:
I currently don't have any other plans for it, so would be in favour of making it private.

Preventing it being called multiple times would definitely be nice, although I don't think it would break the `make_safe` code if it was.

sedders123:
👍  My thoughts on preventing it from being run more than once:

Right now `_setup` just adds to the `BOARDS` list, though it feels like there could potentially be extra setup steps added in the future. As we're making it private my gut says to make it raise an exception if called more than once, to aid in debugging while developing the package, the alternative being silently doing nothing which could be confusing.

I can see two approaches to how we implement this, make it check whether the board is in the `BOARDS` list and throw if it already is, or store the board's setup state in a private attribute and check that before throwing if necessary.

Personally I prefer the second option as it decouples 'setting up' the board from adding it to the `BOARDS` list and also allows for code to check whether the board is setup with clear intnet (`if _board_setup` vs `if self in Board.BOARDS`). Though I'm happy to implement the first option if others feel differently.

Any opinions?

trickeydan:
> Right now `_setup` just adds to the `BOARDS` list, though it feels like there could potentially be extra setup steps added in the future. As we're making it private my gut says to make it raise an exception if called more than once as this would aid in debugging during package development, the alternative being silently doing nothing which could be confusing.

I agree.

> 
> I can see two approaches to how we implement this, make it check whether the board is in the `BOARDS` list and throw if it already is, or store the board's setup state in a private attribute and check that before throwing if necessary.
> 
> Personally I prefer the second option as it decouples 'setting up' the board from adding it to the `BOARDS` list and also allows for code to check whether the board is setup with clear intnet (`if _board_setup` vs `if self in Board.BOARDS`). Though I'm happy to implement the first option if others feel differently.

> Any opinions?

Second option looks good. Let's decouple them so that it doesn't break things in the future.

j5/boards/base.py
```diff
@@ -27,6 +27,7 @@ def __str__(self) -> str:
     def __new__(cls, *args, **kwargs):
         """Ensure any instantiated board is correctly setup."""
         instance = super().__new__(cls)
+        instance._board_setup = False
```
sedders123: Not a huge fan of this monkey patching, but as this needs to be an instance attribute I can't see a way around it. As it is an internal attribute not meant to be used by consumers it makes me feel slightly better about this. Happy to discuss alternative options.

kierdavis:
What is the purpose of the setup method? What does it do that cannot or
should not be done by the constructor?

sedders123:
The setup method is called from the __new__ method, and currently adds the board to a global list of boards for purposes of making them safe at a later time. 

The current approach was taken with the assumption that we may want to do more setup in the future. If there is little to no chance that we will do any more board 'setup'  then it definitely makes sense to ditch the setup method and just add to the list from directly within new.

 In my mind it seems likely that _specific implementations_ of Board could have things that need to be 'setup', but I'm struggling to think of something that all boards will need setup in our base class.

Also one of the reasons it's called from new is that we don't wan't consumers to opt-out of our board list registration, but currently they can do this by overriding the `_setup` method.

I reckon we should just add boards directly in the new method and remove the setup method, though happy to discuss if I've missed a benefit of keeping the method.

trickeydan:
It is preferable to have the board instance added to the `BOARDS` list **not** in the board constructor, as it would be easier for somebody to forget to add it to the list. The `BOARDS` list is require to ensure that the `make_safe` method is called on the board when the code dies.

kierdavis:
I agree with `__new__` being the best place to add the board to `BOARDS`, since it is difficult to override `__new__` without calling the superclass's implementation (unlike when overriding `__init__` or `setup`).

My point was about whether we're expecting `setup` to do other things too e.g. initialising data structures or opening device drivers. It seems to me that these tasks are part of initialisation of the `Board` and hence should be done in the `Board`'s `__init__`. Of course, if these tasks are significant in length it would make sense to move them to another method which is called from `__init__` in order to reduce the complexity of `__init__` - however this should be done on a per-`Board` basis, and the method should be given a more descriptive name than `setup`.

j5/boards/base.py
kierdavis: What's wrong with `board in BOARDS`?

kierdavis: Specifically:

```python
def _setup(self):
  if self not in Board.BOARDS:
    Board.BOARDS.append(self)
  else:
    raise Exception(f"Setup has already been called for board {self.serial}")
```

sedders123:
Based on this discussion I'll remove the `_setup()` method and the associated `_board_setup` attribute, moving the 'board registration' directly into the `__new__` method. 

If I've misunderstood anything let me know, if not I'll implement these changes this evening

j5/boards/base.py
sedders123: Render moot by later discussion, but as mentioned in a previous comment this was about conveying intent clearly.

sedders123:
If there is nothing else for me to do are you OK to merge this @trickeydan?
image

kierdavis:
Who *is* an authorised user? I thought I was (since I've been able to merge PRs into `master` before, e.g. #26), but I don't have the permissions to merge this one it seems.

sedders123:
I have been able to merge in the past, I guess the permissions have been changed

trickeydan:
Erm. You all should be authorised users... I'll go check the settings. I thought I'd put @kierdavis as an org admin

PR #57, opened 2019-02-08
Create a Button Component

### LINKED ISSUE #44: Button Component
Required for #14

trickeydan:
Closes #44

### DIFF
--- a/j5/components/button.py
+++ b/j5/components/button.py
@@ -10,8 +10,8 @@ class ButtonInterface(metaclass=ABCMeta):
     """An interface containing the methods required for a button."""
 
     @abstractmethod
-    def get_led_state(self, board: Board, identifier: int) -> bool:
-        """Set the state of an LED."""
+    def get_button_state(self, board: Board, identifier: int) -> bool:
+        """Set the state of a button."""
         raise NotImplementedError  # pragma: no cover
 
     @abstractmethod
@@ -35,8 +35,8 @@ class Button(Component):
 
     @property
     def is_pressed(self) -> bool:
-        """Get the current pushed state of the LED."""
-        return self._backend.get_led_state(self._board, self._identifier)
+        """Get the current pushed state of the button."""
+        return self._backend.get_button_state(self._board, self._identifier)
 
     def wait_until_button_pressed(self):
         """Halt the program until this button is pushed."""
--- a/j5/components/button.py
+++ b/j5/components/button.py
@@ -1,0 +1,43 @@
+"""Classes for Button."""
+
+from abc import ABCMeta, abstractmethod
+
+from j5.boards import Board
+from j5.components import Component
+
+
+class ButtonInterface(metaclass=ABCMeta):
+    """An interface containing the methods required for a button."""
+
+    @abstractmethod
+    def get_led_state(self, board: Board, identifier: int) -> bool:
+        """Set the state of an LED."""
+        raise NotImplementedError  # pragma: no cover
+
+    @abstractmethod
+    def wait_until_button_pressed(self, board: Board, identifier: int) -> bool:
+        """Halt the program until this button is pushed."""
+        raise NotImplementedError  # pragma: no cover
+
+
+class Button(Component):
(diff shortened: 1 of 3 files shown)


tests/components/test_button.py
```diff
@@ -0,0 +1,81 @@
...
+    end_time = time()
+    time_taken = end_time - start_time
+
+    assert time_taken > 0.2
+    assert time_taken < 2
+
+
+def test_button_interface_class():
+    """Test that the Button Interface class is a ButtonInterface."""
+    assert Button.interface_class() == ButtonInterface
```
kierdavis: Would `is` be more pythonic than `==` here?

tests/components/test_button.py
```diff
@@ -0,0 +1,81 @@
...
+
+def test_button_state():
+    """Test that we can get the state of the button."""
+    button = Button(0, MockButtonBoard(), MockButtonDriver())
+
+    assert not button.is_pressed
+
+
+def test_button_wait_until_pressed():
+    """Test that the button takes at 0.2 secs to be pressed."""
```
kierdavis:
```suggestion
    """Test that the button takes at least 0.2 secs to be pressed."""
```

tests/components/test_button.py
```diff
@@ -0,0 +1,81 @@
...
+    def serial(self) -> str:
+        """The serial number of this board."""
+        return "SERIAL"
+
+    def make_safe(self):
+        """Make this board safe."""
+        pass
+
+    @staticmethod
+    def supported_components() -> List[Type['Component']]:
```
kierdavis: Why is `Component` in quotes here?

Another occurrence on line 45.

tests/components/test_button.py
```diff
@@ -0,0 +1,81 @@
...
+def test_button_instantiation():
+    """Test that we can instantiate a button."""
+    Button(0, MockButtonBoard(), MockButtonDriver())
+
+
+def test_button_state():
+    """Test that we can get the state of the button."""
+    button = Button(0, MockButtonBoard(), MockButtonDriver())
+
+    assert not button.is_pressed
```
kierdavis: Since this function has direct ownership of the `MockButtonDriver`, I think it would make sense to have `MockButtonDriver.get_button_state` return an attribute of the `MockButtonDriver` instance which is set to different values by this function. This would allow the test function to do a better job of verifying that `MockButtonDriver.get_button_state` propagates to `Button.is_pressed` by testing more than one value.

kierdavis: Additionally, part of me wants to change this assertion to `assert button.is_pressed is False` to verify that we're returning a boolean and not anything else that happens to be false-y (e.g. an empty string). I don't know whether this test scenario is useful to consider though. What are your thoughts?

tests/components/test_button.py
trickeydan: No reason.

tests/components/test_button.py
trickeydan: > Additionally, part of me wants to change this assertion to `assert button.is_pressed is False` to verify that we're returning a boolean and not anything else that happens to be false-y (e.g. an empty string). I don't know whether this test scenario is useful to consider though. What are your thoughts?

I feel like this is what a type checker does, but have done it for you anyway.

j5/components/button.py
```diff
@@ -0,0 +1,43 @@
...
+    def interface_class():
+        """Get the interface class that is required to use this component."""
+        return ButtonInterface
+
+    @property
+    def is_pressed(self) -> bool:
+        """Get the current pushed state of the button."""
+        return self._backend.get_button_state(self._board, self._identifier)
+
+    def wait_until_button_pressed(self):
```
kierdavis: Minor nitpick, I'd personally call this `wait_until_pressed`. The word "button" is probably redundant here since the class name already makes it clear that this method is related to a button. I don't mind if you'd rather keep it though.

trickeydan: Agreed.

PR #58, opened 2019-02-08
Add return type to MockLEDBoard

trickeydan:


trickeydan:
I originally imported `Component` for the type checking, but the linter didn't like it.

sedders123:
Yeah, I can't see immediately _why_ though. Maybe because `LED` is a subclass `Component`, it is in the scope of the type checker? (Wild speculation)

sedders123:
Looking into it I think [these docs](https://mypy.readthedocs.io/en/latest/running_mypy.html#following-imports) justify this behaviour. We have the import `LED` and the file containing `LED` imports `Component`. By following the import chain mypy knows about `Component` without needin an explicit import in this file 😃

PR #69, opened 2019-02-12
Tighten up mypy config

### LINKED ISSUE #68: Enforce type hints on all code
Now that the majority of the project has type hints, we should try to keep it that way. We can enforce this by changing our mypy settings to enforce code has type hints.

I propose we enable the following flags, though I'm open to discussion (these are the command line flags but they have equivalent `.cfg` options):
 - --disallow-untyped-defs
 - --disallow-incomplete-defs
 - --no-implicit-optional (this one is more personal preference and less to do with maintaining type coverage)

Docs for the above flags can he found [here](https://mypy.readthedocs.io/en/latest/command_line.html#untyped-definitions-and-calls)

sedders123:
Fixes #68.

Have had to add a base `Interface` class to appease the `disallow_any_generics` flag.

Two flags are currently disabled as they cause errors where we use `atexit` and `__new__`, eager to discuss what we should do here.

Unrelated `pipenv run lint` gives a different result locally to that on CI. Almost certain this is something wrong with my local Windows setup, but will investigate and open new issue if necessary

### DIFF
--- a/j5/backends/base.py
+++ b/j5/backends/base.py
@@ -15,7 +15,7 @@ class BackendMeta(ABCMeta):
 
     """
 
-    def __new__(mcs, name, bases, namespace, **kwargs):
+    def __new__(mcs, name, bases, namespace, **kwargs):  # type:ignore
         """Create a new class object."""
         cls = super().__new__(mcs, name, bases, namespace, **kwargs)  # type: ignore
 
@@ -51,7 +51,7 @@ class Backend(metaclass=BackendMeta):
 
     @property
     @abstractmethod
-    def environment(self):
+    def environment(self) -> 'Environment':
         """Environment the backend belongs too."""
         raise NotImplementedError  # pragma: no cover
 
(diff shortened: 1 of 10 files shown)


trickeydan:
> Two flags are currently disabled as they cause errors where we use `atexit` and `__new__`, eager to discuss what we should do here.

Having a look at this now

kierdavis:
Typechecking is currently failing (probably as a result of the recent merge introducing new untyped code).

```
pipenv run type
j5/components/button.py:32: error: Function is missing a type annotation
j5/components/button.py:41: error: Function is missing a type annotation
Exited with code 1
```

j5/backends/base.py
```diff
@@ -51,7 +51,7 @@ class Backend(metaclass=BackendMeta):
 
     @property
     @abstractmethod
-    def environment(self):
+    def environment(self) -> 'Environment':
```
kierdavis: Why do quotes need to be used here?

sedders123: Because `Environment` hasn't yet been declared when it gets to this line in the file, causing Python to fall over. MyPy will see the strings and understand that it means `Environment` as they look for all classes and types before checking validity

PR #76, opened 2019-02-13
Implement power output component

### LINKED ISSUE #61: Add PowerOutput Component
We need a component to represent the power outputs on the power board.

Another nice things would be a neat way to group them together.

kierdavis:
Models one of the power board's six outputs, including current measurement. Hopefully self-explanatory.

Closes #61.

### DIFF
--- a/tests/components/test_power_output.py
+++ b/tests/components/test_power_output.py
@@ -1,0 +1,80 @@
+"""Tests for the power output classes."""
+from typing import List, Type
+
+from j5.backends import Backend
+from j5.boards import Board
+from j5.components.power_output import PowerOutput, PowerOutputInterface
+
+
+class MockPowerOutputDriver(PowerOutputInterface):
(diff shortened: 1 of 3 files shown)


tests/components/test_power_output.py
```diff
@@ -0,0 +1,80 @@
...
+        """The name of this board."""
+        return "Testing Power Output Board"
+
+    @property
+    def serial(self) -> str:
+        """The serial number of this board."""
+        return "SERIAL"
+
+    @property
+    def supported_components(self) -> List[Type["Component"]]:
```
sedders123: `Component` doesn't have to be in quotes, though I'm happy for them to be left if you'd rather not add an additional import to the file

kierdavis: I copy-pasted this from another test file, so I think I'll leave it as is. We can change them all later if we think that's the right thing to do.

PR #77, opened 2019-02-13
Ensure that Pipfile.lock is kept up to date

kierdavis:
This is done by passing `--deploy` to `pipenv install` during the CircleCI script, which is documented to "abort if the Pipfile.lock is out-of-date, or Python version is wrong". The aim is to ensure
that the versions of Pipfile and Pipfile.lock in the repository are kept consistent.

Tested by adding a new package (`numpy`, arbitrarily) to the list of dependencies in the Pipfile and running a CircleCI build, which results in an error as expected:

    Your Pipfile.lock (e6a0f9) is out of date. Expected: (938333)

Before this commit, the same action would cause `pipenv install` to regenerate Pipfile.lock and install the spurious package - however the version of this package is not locked by the Pipfile.lock
committed in the repository.

I could be a kernel hacker with commit messages like this...

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,7 +6,7 @@ jobs:
     steps:
       - checkout
       - run: pip install pipenv
-      - run: pipenv install --dev
+      - run: pipenv install --deploy --dev
       - run:
           name: Setup Code Climate test-reporter
           command: |


kierdavis:
Thinking about it we should perhaps be using `pipenv sync` instead of `pipenv install`; AFAIK the former ignores the `Pipfile` and just installs what's specified exactly in `Pipfile.lock`. Thoughts?

sedders123:
I reckon it'd be safer to use `pip sync` especially with our current Pipfile setup (wildcards). Even if we had a dependency bot builds would still fail until those PRs were merged in. It also means builds that are repeated will behave exactly the same way whenever they are run.

trickeydan:
Happy for you to do whatever you think is best

kierdavis:
As the general consensus is that `pipenv sync` is preferable, I'm closing this in favour of #88.

PR #93, opened 2019-02-15
[WIP] SR v4 Power Board

### LINKED ISSUE #59: Support SR v4 Power Board
Support SR v4 Power Board.

Prerequisites:
- ~~Button Support~~ - #44 
- ~~Power Output Support~~ - #61
- ~~Battery (voltage / status)~~ - #62
- ~~Piezo~~ - #60

trickeydan:
Closes #59


trickeydan:
Not finished yet obviously, but a review of the current structure would be appreciated before I continue :)

j5/boards/sr/__init__.py
```diff
@@ -0,0 +1 @@
+"""Boards made by Student Robotics."""
```
sedders123: At one point there were discussions of having competition specific implementations in their own packages. This would mean a smaller 'core' j5 and consumers can then specify which specific implementations they wish to include in their projects.

This is not a huge deal right now, but we should decide whether this is something we still want to do and maybe open an issue to track it.

j5/boards/sr/v4/power_board.py
```diff
@@ -0,0 +1,129 @@
...
+PowerOutputGroupIndex = Union[int, PowerOutputPosition]
+
+
+class PowerOutputGroup:
+    """A group of PowerOutputs on the PowerBoard."""
+
+    def __init__(self, backend: Backend, board: Board):
+        self._backend = backend
+        self._board = board
+        self._outputs = [
```
sedders123: Should this be a Dictionary of Position to Output? The code relies on the fact the order of the elements in the list will stay the same. It's easy for a developer to sort a list and not realise they are changing behaviour but harder for them to change the keys in a dictionary and not realise they are doing something dodgy.

Not saying this has to change but those are my immediate thoughts

j5/boards/sr/v4/power_board.py
```diff
@@ -0,0 +1,129 @@
...
+            PowerOutput(n, self._board, cast("PowerOutputInterface", self._backend))
+            for n in range(0, 6)
+        ]
+
+    def power_on(self) -> None:
+        """Set all outputs in the group on."""
+        for output in self._outputs:
+            output.is_enabled = False
```
sedders123: Am I misunderstanding the `is_enabled` flag? In my mind 'turning on' the output would set `is_enabled` to true.

j5/boards/sr/v4/power_board.py
```diff
@@ -0,0 +1,129 @@
...
+        return self._start_button
+
+    @property
+    def battery_sensor(self) -> BatterySensor:
+        """Get the battery sensor."""
+        return self._battery_sensor
+
+    def make_safe(self) -> None:
+        """Make this board safe."""
+        self.outputs.power_off()
```
sedders123: As we have access to `self._output_group` would it be an idea to call `power_off` on that directly? Removes the chance that later down the line we change the `outputs` property and forget to change the `make_safe` functionality

tests/boards/sr/__init__.py
```diff
@@ -0,0 +1 @@
+"""Tests for Student Robptics boards."""
```
sedders123:
```suggestion
"""Tests for Student Robotics boards."""
```

j5/boards/sr/__init__.py
trickeydan: @kierdavis and I discussed this a couple of weeks ago and concluded that we should keep things together for now. In the future, I definitely think that things should be separate though and would actually be happier if they were separated now. Unix philosophy and all that. Besides, the whole point of `j5` is that it is modular.

I have opened #97 to discuss this further.

j5/boards/sr/v4/power_board.py
trickeydan: I definitely would prefer this.

j5/boards/sr/v4/power_board.py
trickeydan: This is in fact a typo

trickeydan: Fixed in a460a75

j5/boards/sr/v4/power_board.py
trickeydan: Fixed in 176195f

j5/boards/sr/v4/power_board.py
```diff
@@ -0,0 +1,126 @@
...
+        PowerOutputInterface,
+        PiezoInterface,
+        BatterySensorInterface,
+        LEDInterface,
+    )
+    from typing import Type  # noqa: F401
+
+
+class PowerOutputPosition(Enum):
+    """A mapping of name to number of the PowerBoard outputs."""
```
kierdavis: I think its worth documenting here that the numbers are the same as are used in the wire protocol, to prevent anyone deciding to renumber the outputs in the future.

trickeydan:
Currently not building due to typing. This is fixed in #111 and this PR is now blocked until that is merged.

trickeydan:
This should fix #59 I hope

kierdavis:
I have a [work-in-progress implementation of a hardware backend](https://github.com/j5api/j5/tree/sr-v4-power-board-backend) for the power board. Should we aim to merge that into this PR so that all the power board stuff comes in at once, or wait for this PR to be merged first? If the latter, I think we're good to go on this PR.

trickeydan:
Let's wait for your backend I think and then we can test the whole stack before merging.

j5/backends/hardware/sr/v4/power_board.py
```diff
@@ -65,6 +66,41 @@ class WriteCommand(NamedTuple):
...
+                usb1.USBErrorOverflow,
+                usb1.USBErrorPipe,
+                usb1.USBErrorInterrupted,
+                usb1.USBErrorNoMem,
+                usb1.USBErrorNotSupported,
+                usb1.USBErrorOther,
+
+        ) as e:
+            name = usb1.libusb1.libusb_error.get(e.value, 'Unknown Error.')
+            raise CommunicationError(f"USB Error({name})") from None
```
kierdavis: Please can we not use `from None` here? It makes it harder to find the underlying cause of the error, which sometimes may require delving into the source of `usb1` if the error messages provided by that library are sufficiently unhelpful.

j5/boards/sr/v4/power_board.py
```diff
@@ -1,6 +1,7 @@
 """Classes for the SR v4 Power Board."""
 
 from enum import Enum
+from functools import wraps
```
kierdavis: Is this import used?

trickeydan: Nope. Hence it fails linting.

trickeydan:
Todo:

[ ]: Run LED should flash on start button wait.
[ ]: Tests to get good coverage.
[ ]: Remove `pipenv run test_hw` - no longer used.
[ ]: Update `setup.py` to use PyUSB
[ ]: Update `shell.nix` to use PyUSB?
[ ]: Stress test it.
[ ]: (Optional): Test on a Windows machine.
[ ]: :shipit:

kierdavis:
Flashing the run LED falls outside the scope of the `wait_until_button_pressed` method of `ButtonInterface`. Somewhat relatedly, `wait_until_button_pressed` doesn't do anything backend-specific - this functionality could easily be implemented at a higher level (e.g. the Board or Robot level).

I'm tempted to say that both of these issues could be solved by binning `ButtonInterface.wait_until_button_pressed` and creating a `PowerBoard.wait_until_start_button_pressed` that does both the polling and the flashing.

kierdavis:
Seeing as this PR is now monstrously large and is pulling in all sorts of unrelated changes just to keep up with master, I recommend we merge it now and spin out anything else we want to do on this topic into separate issues.

Issue #79, opened 2019-02-13
Specify minimum supported Python version


kierdavis:
We should specify what minimum version of Python is required to run j5. Things to consider:

* We likely want to use recent additions to the language, such as:
  * Predictably ordered `dict`s, introduced in 3.7. A concrete use of this feature is highlighted in [this comment](https://github.com/j5api/j5/pull/74#discussion_r256566865).
  * f-strings, introduced in 3.6.
* The Linux distributions on which j5 may be required to run might not provide recent Python versions.
* Read The Docs apparently doesn't support Python 3.7 (#49).

sedders123:
Read The Docs [seem to say](https://github.com/rtfd/readthedocs-docker-images/blame/master/README.rst#L30) they support Python 3.7. Was there an error that Read the Docs gave when we tried to use it?

kierdavis:
Not sure. #46 also looks relevant.

@trickeydan please can you elaborate?

trickeydan:
Read the docs claim they support 3.7, but in reality it doesn't work. That version is only relevant for the building of the docs as far as I can tell.

In terms of distro support, I would be happy to use backports. I think that it is super important to support the latest versions of things whenever possible. I can definitely see a use case for supporting old versions of things if we are aware of a kit that uses it (e.g we don't want to upgrade to 3.8 at the start of the 2020 comp cycle.). I think that we should discuss a versioning mechanism in a separate ticket.

mildlyincompetent:
The pip / pipenv configuration currently specifies 3.7, so the repo as a whole is not 3.6-compatible. If we do decide to officially support 3.6 we should ensure that we update the configs to reflect this.

kierdavis:
My preference is to stay compatible with 3.6 *for the time being*, to encourage wider adoption of the library. I don't see a problem in increasing the minimum required version when we find we need certain features (or at least, they would make our lives significantly easier), but I don't think its necessary to throw this away *just* yet.

Having said that, our CI runs 3.7 and as @mildlyincompetent points out our Pipfile requires 3.7. I don't even know if the code is compatible with 3.6 at this precise moment. The minimum effort path is thus to stay on 3.7, which I am not opposed to.

trickeydan:
To my knowledge, the codebase is entirely compatible with 3.6.
**However**, I would rather support as few versions of python as possible, preferably the most recent stable versions, which happen to be 3.7. When 3.8 is released, we will continue to support 3.7 for some period of time.

[...]

PR #94, opened 2019-02-15
Add Nix environment definition

kierdavis:
This enables [Nix](https://nixos.org/nix/) users to simply run `nix-shell` in the repository root to obtain a shell with python and pipenv (and anything else we may want to add in the future) available.

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -1,0 +1,16 @@
+{ pkgs ? import <nixpkgs> {} }:
+
+# Note that as of 2019-01-21, pipenv is currently broken in the NixOS 18.09 release.
+# https://github.com/NixOS/nixpkgs/issues/51970
+# To work around, invoke with:
+# nix-shell -E 'import ./shell.nix { pkgs = import (fetchTarball https://nixos.org/channels/nixpkgs-unstable/nixexprs.tar.xz) {}; }'
+
+with pkgs;
+
+stdenv.mkDerivation {
+  name = "j5-dev-env";
+  buildInputs = [
+    pipenv
+    python3
+  ];
+}
--- a/shell.nix
+++ b/shell.nix
@@ -1,6 +1,13 @@
+# A nixpkgs overlay to set python3Packages to python37Packages (the default is python36Packages).
+let
+  python37ByDefault = self: super: {
+    python3Packages = self.python37Packages;
+  };
+in
+
 {
   pkgsSrc ? <nixpkgs>,
-  pkgs ? import pkgsSrc {}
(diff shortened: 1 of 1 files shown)


mildlyincompetent:
I recall commenting on a similar PR that this is nasty and should not go in the repo, however as I am now a NixOS user, I think this is a great idea and should go in all repos.

trickeydan:
> I recall commenting on a similar PR that this is nasty and should not go in the repo, however as I am now a NixOS user, I think this is a great idea and should go in all repos.

Well I'm strongly considering building the new Pi OS on nix, so this is very useful.

mildlyincompetent:
Huh, on my system the `python3` requirement was satisfied using python 3.6, which doesn't work as we require 3.7. Is it possible to specify this more precisely?

kierdavis:
@mildlyincompetent it is possible (`python3` -> `python37`).

I thought we were still compatible with 3.6? We really need to make a decision on #79.

mildlyincompetent:
The pip / pipenv config isn't, so the repo isn't 3.6 compatible currently, though this may be subject to change once we work out what exactly we're supporting.

kierdavis:
Regardless, all evidence suggests that 3.7 is the preferred version for development, so I'll specify this in the `shell.nix`.

kierdavis:
Ugh. As a result of Nix's default python3 version being 3.6 (on the 18.09 release, haven't checked unstable), one must go out of their way to override `python3Packages = python37Packages`. Hydra doesn't appear to build python packages against any python version other than the default, so using a non-default python version here forces pipenv and its dependencies to be built locally. Luckily this is all automated but its a bit annoying for `nix-shell` to now take 10 minutes to set the environment up.

mildlyincompetent:
Hey at least we're not attempting to `emerge`, could be a worse distro. Does this process only take forever once (since I assume its results live in the nix store), or does it take ages every time? If its the former I would say it is grudgingly acceptable.

Unstable uses 3.7 I believe judging by the random docker container Dan just downloaded, and appears to have binaries for 3.7.

kierdavis:
Yes, results will persist in the store, so it'll only take forever once (unless you garbage collect the store automatically, in which case it'll take forever once per week).

kierdavis:
Thinking about it the point about nix-shell taking a long time is moot. [pipenv is broken on the current stable channel](https://github.com/NixOS/nixpkgs/issues/51970), so one must use unstable in the meantime. The next release will almost certainly make python 3.7 the default.

PR #95, opened 2019-02-15
Acknowledge all contributors to the project

kierdavis:
As a static list of contributors would quickly become outdated,
this commit introduces a script which generates a list from the
git commit history. A [mailmap][] file is used to normalise
author names and email addresses.

This also changes the copyright declaration in the license file to refer generally to "j5 contributors". Question: should the LICENSE file explicitly state that the copyright is owned by the individuals named in CONTRIBUTORS, or is that unnecessary?

[mailmap]: https://github.com/git/git/blob/master/Documentation/mailmap.txt


trickeydan:
Tyler has raised some concerns about having a static / generated file in version control. How is this updated? Is it automatic?
It's worth noting that the CLA requires contributors to give up their claim to copyright to the project. So none of us own the copyright to the project I think technically. I personally would rather that the project belonged to itself but I don't know how well that has worked for others in the past.

I'll do some research into models of open source projects.

kierdavis:
I'm aware that having to keep `CONTRIBUTORS` up to date is rather unclean. Possible solutions I can think of:

* Use a bot to automatically update this file whenever a PR is merged.
* Use a bot to prevent PRs being merged unless a human has updated the file.
* Forgo the `CONTRIBUTORS` file and ask users to just run `CONTRIBUTIONS.gen` instead.

trickeydan:
So it's worth noting that CLA Assistant keeps a list of all contributors also, along with their GitHub email as a means to contact them. Contributors also have to agree to the CLA in order for a PR to be merged, and that is the only way to get commits on `master`. 
I think I would be happy with a comment in the `README.md` stating how to execute `CONTRIBUTORS.gen`. Happy to discuss further later.

@mildlyincompetent I would value your opinion here :)

mildlyincompetent:
The contributors.gen solution seems appropriate to me - a CONTRIBUTORS file is not needed from a legal perspective if all contributors have granted their copyright to the project (although to my knowledge the project is not really a legal entity that can own IP right now, so that may be another can of spaghetti hoops).

Furthermore, as we are considering the master branch to be essentially append-only (excluding blatant abuse), a CONTRIBUTORS file is essentially duplicating data contained in the git history, so unless it's serving some specific purpose I don't really see the need for it. Obviously it's nice for people to have their contributions acknowledged but I feel like GitHub does a reasonably good job of this anyway.

README.md
```diff
@@ -73,5 +73,13 @@ As this library has to support multiple hardwares, with multiple backends and st
 
 `j5` is being developed by a group of volunteers at the University of Southampton. We welcome contributions and reside in a channel on the SourceBots Slack.
 
+## Contributions
+
+This project is released under the MIT Licence. For more information, please see `LICENCE`.
```
kierdavis:
```suggestion
This project is released under the MIT Licence. For more information, please see `LICENSE`.
```

Issue #99, opened 2019-02-17
Insufficient Unit tests on Piezo


trickeydan:
We need to test if the buzz method is called, and if the frequency is out of bounds.

trickeydan:
@HU90m I saw you doing this last night, so have assigned it to you

trickeydan:
This was done in #119

PR #119, opened 2019-02-20
Input validation for piezo's buzz method and improved test coverage

HU90m:
Solves issues #99 and #100 .


j5/components/piezo.py
```diff
@@ -16,18 +16,18 @@ class Note(Enum):
     notation and their related frequencies in Hz.
     """
 
-    C6 = 1046
-    D6 = 1174
-    E6 = 1318
-    F6 = 1396
-    G6 = 1567
+    C6 = 1047
```
trickeydan: I initially wondered why these were changed, but I now see that the previous values were not correct.

j5/components/piezo.py
```diff
@@ -61,4 +61,14 @@ def interface_class() -> Type[PiezoInterface]:
     def buzz(self, board: Board, identifier: int,
              duration: timedelta, pitch: Pitch) -> None:
         """Queue a note to be played."""
-        self._backend.buzz(board, identifier, duration, pitch)
+        if type(pitch) is int:
+            frequency = pitch
+        elif type(pitch) is Note:
+            frequency = pitch.value  # type: ignore
```
trickeydan: I worked with @HU90m to try and get these types to work with no success. If we can't fix them before merging, I suggest opening an issue so we don't forget about them.

@sedders123 can you help here?

sedders123: Unfortunately `enum` values are a [bit of a pain](https://github.com/python/mypy/issues/3546) in mypy, though there are [plans](https://github.com/python/mypy/issues/3541) to try and improve it with plugins. In the meantime would a `cast` to type `int` work here?

trickeydan: We tried `cast` earlier and unfortunately it didn't work.

trickeydan: Aha. We should try `IntEnum` as it has the type of `value` defined as `int` in https://github.com/python/typeshed/blob/master/stdlib/3/enum.pyi#L35

sedders123:
```python
def buzz(self, board: Board, identifier: int,
             duration: timedelta, pitch: Pitch) -> None:
        """Queue a note to be played."""
        if type(pitch) is int:
            frequency = cast(int, pitch)
        elif type(pitch) is Note:
            frequency = cast(Enum, pitch).value
        else:
            raise TypeError

        if frequency < 0:
            raise ValueError
        else:
            self._backend.buzz(board, identifier, duration, pitch)
```

The above code seems to pass the linter. Had to cast on both `int` case and `Note` case.

trickeydan: Interesting. I couldn't get the above to work.

My solution is PR to this branch #121.

sedders123: 👍 Like the look of #121

j5/components/piezo.py
```diff
@@ -61,4 +61,14 @@ def interface_class() -> Type[PiezoInterface]:
     def buzz(self, board: Board, identifier: int,
              duration: timedelta, pitch: Pitch) -> None:
         """Queue a note to be played."""
-        self._backend.buzz(board, identifier, duration, pitch)
+        if isinstance(pitch, int):
+            frequency = pitch
+        elif isinstance(pitch, Note):
+            frequency = pitch.value
+        else:
+            raise TypeError
```
sedders123:
```suggestion
            raise TypeError("Pitch must be of either type int or Note")
```

j5/components/piezo.py
```diff
@@ -61,4 +61,14 @@ def interface_class() -> Type[PiezoInterface]:
...
+
+        if frequency < 0:
+            raise ValueError
```
sedders123:
```suggestion
            raise ValueError("Pitch must be greater than zero")
```

j5/components/piezo.py
```diff
@@ -61,4 +61,14 @@ def interface_class() -> Type[PiezoInterface]:
...
+        else:
+            self._backend.buzz(board, identifier, duration, pitch)
```
kierdavis: I would prefer for the `frequency` to be passed to `self._backend.buzz` rather than `pitch` (it seems like currently we just discard `frequency` after validating it?). However since making this change affects all backends, I think it's outside the scope of this PR. I have created #122 to track this.

tests/components/test_piezo.py
```diff
@@ -53,3 +55,25 @@ def test_piezo_interface_implementation():
...
+
+
+def test_piezo_interface_class_method():
+    """Tests piezo's interface_class method."""
+    piezo = Piezo(0, MockPiezoBoard(), MockPiezoDriver())
+    assert piezo.interface_class() is PiezoInterface
+
+
+def test_piezo_buzz_method():
+    """Tests piezo's buzz method."""
```
kierdavis:
```suggestion
    """Tests piezo's buzz method's input validation."""
```

This test does not check that the backend's buzz method is actually called with the correct parameters, so the name and docstring of the test is a bit misleading. I would be in favour of merging this test into `test_piezo_buzz_invalid_value`

PR #123, opened 2019-02-21
Make Environment.get_backend return a class

### LINKED ISSUE #118: Backend-to-Board correspondence
Are we expecting there to be one instance of `Backend` corresponding to each instance of `Board`, or for there to be a singleton instance of `Backend` that operates on any `Board`? To put it another way, should `Backend` objects store board-specific information (e.g. USB device handle) in their attributes, or should this information be stored on the `Board`?

If the former, we need to review how we create and use `Backend`s. My understanding is that our flow currently looks like this (please correct me if I'm wrong):
* `Robot.__init__` calls `Environment.get_backend(board_class)`, obtaining `backend`, a new instance of `backend_class`
* `Robot.__init__` creates a `BoardGroup`, passing `backend`
* `BoardGroup.__init__` calls `BoardGroup.update_boards`, which calls `backend.discover()`
* `backend.discover` creates and returns a number of `board_class` instances, passing the serial number and `self.environment` to the constructor. It is undefined where backend-specific data about the board is kept
* `Board.__init__` calls `Environment.get_backend(self.__class__)`, creating another instance of `backend_class` on which component operations for that board will be performed
* `BoardGroup.update_boards` stores the collection of `Board` instances

I would expect the flow to look something more like:
* `Robot.__init__` calls `Environment.get_backend(board_class)`, obtaining `backend`, **the only instance** of `backend_class` (using the [singleton pattern](https://en.wikipedia.org/wiki/Singleton_pattern))
* `Robot.__init__` creates a `BoardGroup`, passing the `backend`
* `BoardGroup.__init__` calls `BoardGroup.update_boards`, which calls `backend.discover()`
* `backend.discover` creates and returns a number of `board_class` instances, passing the serial number, **the backend** to the constructor. Backend-specific data about the board is held by the backend instance, in a dict keyed by board instance.
* `BoardGroup.update_boards` stores the collection of `board_class` instances

However, if we want to have one `Backend` per `Board`, the flow should probably look something like this:
* `Robot.__init__` calls `Environment.get_backend(board_class)`, obtaining **the corresponding `Backend` class**
* `Robot.__init__` creates a `BoardGroup`, passing the **backend class**
* `BoardGroup.__init__` calls `BoardGroup.update_boards`, which calls `backend_class.discover()`
* `backend_class.discover` creates a number of `backend_class` instances, which can hold arbitrary backend-specific data about the corresponding board
* `backend_class.discover` wraps each `backend_class` instance with a `board_class` instance.
* `BoardGroup.update_boards` stores the collection of `board_class` instances

What are your thoughts on these approaches?

kierdavis:
Resolves #118.

### DIFF
--- a/j5/backends/base.py
+++ b/j5/backends/base.py
@@ -86,9 +86,9 @@ class Environment:
         """Register a new backend with this Backend Group."""
         self.board_backend_mapping[board] = backend
 
-    def get_backend(self, board: 'Type[Board]') -> Backend:
+    def get_backend(self, board: 'Type[Board]') -> Type[Backend]:
         """Get the backend for a board."""
         if board not in self.supported_boards:
             raise NotImplementedError(f"The {str(self)} does not support {str(board)}")
 
-        return self.board_backend_mapping[board]()
+        return self.board_backend_mapping[board]
--- a/j5/boards/j5/demo.py
+++ b/j5/boards/j5/demo.py
@@ -17,7 +17,7 @@ class DemoBoard(Board):
 
     def __init__(self, serial: str, environment: Environment):
         self._environment = environment
-        self._backend = environment.get_backend(self.__class__)
+        self._backend = environment.get_backend(self.__class__)()
         self._serial = serial
         self._leds = [LED(n, self, cast('LEDInterface', self._backend))
                       for n in range(0, 3)]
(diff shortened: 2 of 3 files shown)


trickeydan:
Worth noting that this will break #93 when merged.

Issue #80, opened 2019-02-14
Document Versioning Mechanism


trickeydan:
We should decide on the versioning system for `j5`.

Relevant material: [PEP440](https://www.python.org/dev/peps/pep-0440/), [PEP396](https://www.python.org/dev/peps/pep-0396/), [SemVer](https://semver.org/).

We need to bear in mind that we may have to maintain multiple major versions at the same time. For example, if the SR competition has deployed version 4.0.1 to kits and we need to deploy a bug fix.

trickeydan:
Also suggested: https://calver.org/

trickeydan:
This has been decided as SemVer.

PR #129, opened 2019-02-22
Consolidate j5 version

trickeydan:
I have created a `j5.VERSION` that is used for docs and packaging. Additionally, this could be used in code to have different behaviour if we need to do so.


sedders123:
Should this be `__version__` or are we not following [PEP 396](https://www.python.org/dev/peps/pep-0396/)? (Not saying we have to, just wondering if there was a reason)

trickeydan:
I was not aware of PEP 396. I will update this.

trickeydan:
Also note: resolves #80

trickeydan:
This is now failing on a lack of Type Defininitions for `packaging`. Any suggestions on how we solve this? I'm not sure if we really should bother making a `.pyi` for it, somewhat out of scope.

kierdavis:
We could add a hint to ignore typing on the offending import.

However, is `packaging.Version` necessary, especially since we immediately convert back to a string? Unless I'm missing something it doesn't seem like introducing a new dependency for this is justified.

trickeydan:
> We could add a hint to ignore typing on the offending import.
> 
> However, is `packaging.Version` necessary, especially since we immediately convert back to a string? Unless I'm missing something it doesn't seem like introducing a new dependency for this is justified.

I made use of `packaging.Version` as it ensures that the version are compatible for PEP440 and it's a requirement of `setuptools` anyway. I'm happy to remove this if you'd like. It also make it possible to dynamically find the short version number that Sphinx wants.

kierdavis:
I can see the benefit if the `Version` object was exposed directly, but if the only purpose is validating the string then that might as well be done once by a human during code review rather than every time the library is loaded. I don't feel that strongly about it though, you're welcome to keep it (so long as the typing gets fixed).

trickeydan:
I think I'm probably in favour of removing it now, especially considering that the types on it are broken.

Issue #120, opened 2019-02-20
Cross-Platform Compatibility


trickeydan:
I'm currently working on #3 and I have a choice of implementations, one of which I would prefer but would also prevent the library working on Windows. I think it is really important to try and keep this cross platform if possible. Currently there is nothing Unix-specific.

I would be interested in the views of others.

sedders123:
I think that we should strive for cross-platform compatibility until we reach a problem that has no other solution but to support just the one platform (or the multiplatform solution requires *substantially* more work).

According to this [site](https://netmarketshare.com/operating-system-market-share.aspx?options=%7B%22filter%22%3A%7B%22%24and%22%3A%5B%7B%22deviceType%22%3A%7B%22%24in%22%3A%5B%22Desktop%2Flaptop%22%5D%7D%7D%5D%7D%2C%22dateLabel%22%3A%22Trend%22%2C%22attributes%22%3A%22share%22%2C%22group%22%3A%22platform%22%2C%22sort%22%3A%7B%22share%22%3A-1%7D%2C%22id%22%3A%22platformsDesktop%22%2C%22dateInterval%22%3A%22Monthly%22%2C%22dateStart%22%3A%222018-02%22%2C%22dateEnd%22%3A%222019-01%22%2C%22segments%22%3A%22-1000%22%7D) Windows has an 87% market share, and while I'm sure in tech circles that number changes it seems crazy to me that we should alienate such a large contributor pool when we don't have to.

I fully understand there are some things that can't be done cross platform, but when we get to a case of this I reckon we should clearly document it, try to make sure that it doesn't make the *whole* project unusable, and try to point out suitable workarounds if there are any (compat libraries, instructions for Windows Sub System for Linux etc.)

kierdavis:
Also note that a j5-based simulator must be able to run on Windows.

My preference would be for a solution that works on all platforms, but choosing between a Unix-specific solution and a Windows-specific solution at runtime would be acceptable.

trickeydan:
I think this issue has been resolved at this point and I shall be implementing a cross-platform solution.

PR #136, opened 2019-02-24
Add lock to BaseRobot to prevent multiple robot processes.

### LINKED ISSUE #3: [FEATURE] Locks
We need to have a lock to prevent multiple robot processes. Suggestions are to use `/var/run`, but that does lock it down to be platform specific which we do not want.

trickeydan:
Resolves #3 

In previous APIs, this has been implemented using a lock file in `/tmp`. For this version it was suggested that we put a PID file in `/var/run`, which was an approach that I initially approved of.

However, up to this point and as confirmed by #120, we are ensuring the cross-platform compatibility of `j5` and I'm sure you're aware that Windows doesn't have `/var/run`. The closest equivalent is called a Mutex and has very poor support on Unix systems.

The solution that I have implemented in this pull request takes leafs out of the book of enterprise Java and isn't exactly what I would do in an ideal world.

Binding to a TCP port on the host is kept to a single process at the kernel level, which means that it is harder to override (e.g no deleting of the PID file.). Additionally, in the case of a crash, the TCP port is automatically closed as we are no longer binding to it.

I am still open to alternative solutions to this and am looking forward to hearing back from you all.


tests/test_stack.py
```diff
@@ -19,3 +22,13 @@ def test_led():
 
     for led in r.demo_board.leds:
         led.state = True
+
+
+def test_robot_lock():
+    """Test that we cannot have more than one Robot object."""
+    r1 = Robot()
+
+    assert r1._lock
```
RealOrangeOne: What exactly is this asserting. It would be better to assert that the address bound to is what you were expecting.

RealOrangeOne:
Whilst I also think that using a socket for this kind of thing is really weird, i've spent 20 minutes looking around for a cross-platform alternative, and came up empty. Although, almost any other application which needs an inter-process lock uses PID files, and I suspect there's a good reason for that! 

I'd also suggest using a much higher, less _l33t_ port. `j5` in ascii is `10653`, which would be a much better fitting port. 

(Also I agree with @sedders123 that locking should not be a public method)

tests/test_stack.py
trickeydan: Fair point. Significantly improved.

kierdavis:
I like this.

PR #138, opened 2019-02-25
GPIO Pin Component

### LINKED ISSUE #43: Create GPIOPin Component

trickeydan:
Resolves #43


j5/components/base.py
```diff
@@ -16,3 +16,9 @@ class Component(metaclass=ABCMeta):
     def interface_class() -> Type[Interface]:
         """Get the interface class that is required to use this component."""
         raise NotImplementedError  # pragma: no cover
+
+
+class NotSupportedByHardware(Exception):
```
sedders123: I think this is is an error and as such such be [suffixed with Error](https://www.python.org/dev/peps/pep-0008/#exception-names), however I don't have a super strong opinion on it if you or others disagree.

sedders123: Also, a question that's outside of the scope of this PR, should we have a base J5Exception that we use as our base Exception class, so that consumers can optionally just catch any of our stuff throwing?

j5/components/gpio_pin.py
```diff
@@ -0,0 +1,191 @@
...
+
+from abc import abstractmethod
+from enum import IntEnum
+from typing import List, Optional, Type
+
+from j5.boards import Board
+from j5.components import Component, Interface, NotSupportedByHardware
+
+
+class BadGPIOPinMode(Exception):
```
sedders123: Again think this should probably have an Error suffix

j5/components/gpio_pin.py
```diff
@@ -0,0 +1,191 @@
...
+    """The pin is not in the correct mode."""
+
+    pass
+
+
+class GPIOPinMode(IntEnum):
+    """Hardware modes that a GPIO pin can be set to."""
+
+    DIGITAL_INPUT = 0  # The digital state of the pin can be read
```
sedders123: Changing these to `#:` will mean that Sphinx will use these as descriptors in the generated docs

j5/components/gpio_pin.py
```diff
@@ -0,0 +1,191 @@
...
+        ])
+        if new_value < 0 or new_value > 1:
+            raise ValueError("An analogue pin value must be between 0 and 1.")
+
+        if self.mode is GPIOPinMode.ANALOGUE_OUTPUT:
+            self._backend.write_gpio_pin_dac_value(
+                self._board,
+                self._identifier,
+                new_value,
+            )
```
sedders123: Think you need a `return` at the end of this if check, otherwise we'll still execute the PWM specific code.

tests/components/test_gpio_pin.py
```diff
@@ -0,0 +1,358 @@
...
+        return 0.6
+
+    def write_gpio_pin_dac_value(
+            self,
+            board: Board,
+            identifier: int,
+            scaled_value: float,
+    ) -> None:
+        """Write a scaled analogue value to the DAC on the GPIO pin."""
+        pass
```
sedders123: Are these (this method and `write_gpio_pin_pwm_value`) methods not being implemented for testing intentionally? Fine if they are, just want to make sure they weren't accidentally missed.

trickeydan: This is as intended. I have ideas about how to make unit testing better, but cannot do that until #93 is finished.

j5/components/gpio_pin.py
kierdavis: Why not simply use an `else` block here?

sedders123: 👍 Or that, either works. `else` is probably more explicit for the next dev who finds it.

j5/components/base.py
trickeydan: I think that's a good idea, and have opened #142 to track it

j5/components/gpio_pin.py
```diff
@@ -0,0 +1,194 @@
...
+        raise NotImplementedError  # pragma: nocover
+
+    @abstractmethod
+    def read_gpio_pin_digital_state(self, board: Board, identifier: int) -> bool:
+        """Read the digital state of the GPIO pin."""
+        raise NotImplementedError  # pragma: nocover
+
+    @abstractmethod
+    def read_gpio_pin_analogue_value(self, board: Board, identifier: int) -> float:
+        """Read the scaled analogue value of the GPIO pin."""
```
kierdavis: What is a "scaled analogue value" in this context?

trickeydan: A float between `0` and `1` that is scaled such that `scaled_value * max_voltage = current_voltage`

trickeydan: I agree that more detailed docs are required for competitor purposes.

PR #141, opened 2019-02-26
Add firmware_version to Board

trickeydan:
Required for #93


trickeydan:
@mildlyincompetent the same board in the API can have different values dependinf on the physically attached board, it is determined by a function in the `Backend`, which should query the board and ask for it's firmware version. For the Demo and Mock Boards I have them all return `None` as they technically do not have a firmware version.

Issue #143, opened 2019-02-26
Define __reverse__ for `Note`


trickeydan:
I think this is a nice addition for playing scales on the `Piezo`

PR #150, opened 2019-03-02
Note reversed

Skybound1:
This completes Issue #143


Skybound1:
The type checker will probably complain about `__members__` being undefined, but this appears to be a bug with the type checker as `__members__` was introduced in py 3.6 but that still doesn't recognise it. 

https://github.com/python/typeshed/issues/1590, following the github issue links leads to the open https://github.com/python/typeshed/issues/1595

trickeydan:
Let's ignore the types on this and add an inline comment with a link to the upstream issue and an issue for us so that we don't forget and can use the upstream fix.

Issue #147, opened 2019-03-02
Piezo: Check for duration < 0


trickeydan:
Currently, there is no check for this.

Issue #122, opened 2019-02-21
PiezoInterface should operate on frequencies, not pitches


kierdavis:
I think it makes sense to pass the frequency to `PiezoInterface.buzz` rather than the pitch, because:

* pitch is a user-facing concept while frequency is a hardware-facing concept
* this would allow the pitch-to-frequency conversion code introduced in #119 to be reused between backends

PR #151, opened 2019-03-02
added input verification to duration parameter of piezo's buzz method

HU90m:
This fixes issues #147 and #122
I also increased the test coverage to 100%.


kierdavis:
Thanks for looking at this! Unfortunately, I think I should have been clearer in #122. I was only intending for the PiezoInterface and its related backends to change, not the Piezo components, and without removing the `Pitch` type.

The aim is that the user passes a `Pitch` (i.e. a note or frequency) to `Piezo.buzz`. `Piezo.buzz` does some validation on these values, and if a note was passed it is converted to a frequency (the code for this already exists, we're just currently throwing the frequency away). This frequency is then passed to `PiezoInterface.buzz`. The end result is that note-to-frequency conversion only happens once (inside `Piezo.buzz` rather than in every backend), and `PiezoInterface` and backends only over have to deal with frequencies.

j5/components/piezo.py
```diff
@@ -67,7 +67,7 @@ def interface_class() -> Type[PiezoInterface]:
 
     def buzz(self, duration: timedelta, pitch: Pitch) -> None:
         """Queue a note to be played."""
-        if type(pitch) is not int and type(pitch) is not Note:
+        if not (type(pitch) is int or type(pitch) is Note):
```
trickeydan: The previous formatting of this is more pythonic.

The bot wants this method split up. I'd recommend creating a static `pitch_to_int` or `validate_pitch` method. Perhaps even both.

j5/components/piezo.py
```diff
@@ -67,8 +67,15 @@ def interface_class() -> Type[PiezoInterface]:
 
     def buzz(self, duration: timedelta, pitch: Pitch) -> None:
         """Queue a note to be played."""
-        if not (type(pitch) is int or type(pitch) is Note):
+        if type(pitch) is int:
+            pass
+        elif type(pitch) is Note:
```
trickeydan: See other comment about complexity

j5/components/piezo.py
HU90m: Ah didn't think of that, thank you!

HU90m: Sorry, I didn't read this properly first time round. I have now done this.

j5/components/piezo.py
```diff
@@ -67,14 +67,21 @@ def interface_class() -> Type[PiezoInterface]:
...
             raise ValueError("Pitch must be greater than zero")
+        elif duration < timedelta(seconds=0):
+            raise ValueError("Duration must be greater than zero")
         else:
             self._backend.buzz(self._board, self._identifier, duration, pitch)
+
+    @staticmethod
+    def verify_pitch(pitch: Pitch) -> bool:
+        """Verifies a pitch is either a int or Note."""
+        pitch_is_not_an_int = type(pitch) is not int
```
HU90m: Are the following variables required/necessary?
```
        pitch_is_not_an_int = type(pitch) is not int
        pitch_is_not_a_Note = type(pitch) is not Note
```

j5/components/piezo.py
```diff
@@ -67,14 +67,21 @@ def interface_class() -> Type[PiezoInterface]:
...
-        if frequency < 0:
+        if pitch < 0:
             raise ValueError("Pitch must be greater than zero")
+        elif duration < timedelta(seconds=0):
+            raise ValueError("Duration must be greater than zero")
         else:
             self._backend.buzz(self._board, self._identifier, duration, pitch)
+
+    @staticmethod
+    def verify_pitch(pitch: Pitch) -> bool:
```
trickeydan:
```suggestion
def verify_pitch(pitch: Pitch) -> None:
        """Verify that a pitch is valid."""
        # Verify that the type is correct.
        pitch_is_int = type(pitch) is  int
        pitch_is_note = type(pitch) is Note
        if not (pitch_is_int or pitch_is_note):
            raise TypeError("Pitch must be int or Note")
       
       # Verify the length of the pitch is non-zero
       if pitch < 0:
           raise ValueError("Frequency must be greater than zero")
           
@staticmethod
def verify_duration(duration: timedelta):
    """Verify that a duration is valid."""
    if not instance(duration, timedelta):
        raise TypeError("Duration must be of type datetime.timedelta")
    if duration < timedelta(seconds=0):
        raise ValueError("Duration must be greater than or equal to zero.")
```

trickeydan: You will need to adjust your other code to use these:
e.g
```
verify_pitch(pitch)
verify_duration(duration)
self._backend.buzz(self._board, self._identifier, duration, pitch)
```

PR #152, opened 2019-03-02
Replace libusb1 with PyUSB

trickeydan:


trickeydan:
Actually, I need to make some type stubs.

j5/backends/hardware/sr/v4/power_board.py
```diff
@@ -65,6 +65,8 @@ class WriteCommand(NamedTuple):
 CMD_WRITE_ERRORLED = WriteCommand(7)
 CMD_WRITE_PIEZO = WriteCommand(8)
 
+usb._objfinalizer._AutoFinalizedObjectBase._do_finalize_object = lambda x: None
```
kierdavis: I agree that this arcane hack is justified, but please can we document why its necessary and why a cleaner approach cannot be used? For the benefit of others reading the code, and for our own future reference.

j5/backends/hardware/sr/v4/power_board.py
```diff
@@ -143,7 +130,8 @@ def __init__(self, usb_device: usb1.USBDevice):
 
     @handle_usb_error
     def _read(self, command: ReadCommand) -> bytes:
-        return self._usb_handle.controlRead(0x80, 64, 0, command.code, command.data_len)
+        request_type = (0x80 & ~0x80) | 0x80
```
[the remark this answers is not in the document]
trickeydan: From https://github.com/vpelletier/python-libusb1/blob/a90474b3388c3e9c6fb8788f2fdcb56980862e6b/usb1/__init__.py#L1497

j5/backends/hardware/sr/v4/power_board.py
trickeydan: Sure.

j5/backends/hardware/sr/v4/power_board.py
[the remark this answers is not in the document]
trickeydan: Happy to swap these out. I actually have little idea what it does, but it's some kind of bitmasking.

Maybe we should name the magic numbers?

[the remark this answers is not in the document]
kierdavis: `request_type` is the `bmRequestType` of a setup packet (part of a control transfer), details of which can be found [here](https://www.beyondlogic.org/usbnutshell/usb6.shtml).

The bit-manipulation ops in the code you linked look like they're there to ensure the direction bit (bit 7) is set appropriately when the direction of the control transfer is known (i.e. when the user has chosen to call `controlWrite` or `controlRead`). As a result, `controlRead(0x00, ...)` doesn't accidentally initiate a write transfer, because the request type `0x00` is corrected to `0x80`.

I don't think this needs to be explicit here. I can see some benefit in naming the constant instead, e.g.

```python
CONTROL_REQUEST_TYPE_WRITE = 0x00
CONTROL_REQUEST_TYPE_READ = 0x80
```

but given that these names don't really add any extra information about what the value of the constant means, and that making this change would isolate the value of the constant from the place where its meaning is relevant (making it a bit harder to figure out what the code is doing), I don't think naming the constants would be a good idea.

stubs/usb.pyi
```diff
@@ -0,0 +1,37 @@
+"""
+Type stubs for PyUSB.
+
+Note that stubs are only written for the parts that we use.
+"""
+
+from typing import Generator, Optional
+
+class Configuration:
+    def ___init__(self, device: Device, configuration: int = 0) -> None: ...
```
kierdavis: There's a lot of definitions in this file that aren't used (as evidenced by the fact that the name of this function contains five underscores and yet it hasn't caused any typechecking errors). I would recommend removing them, so that we're not maintaining things which are not used.

Specifically, I think the following can be removed from this type stub file:

- The `Configuration` class
- `Device.product` and `Device.manufacturer`
- `Device.set_configuration` and `Device.get_active_configuration`
- `Device.reset`

Yes, some of these are useful and we might use them some day. We're not using them right now though, and its easy enough to add solely the required ones back in if/when it turns out we do need them.

stubs/usb/core.pyi
```diff
@@ -0,0 +1,40 @@
+"""
+Type stubs for usb.core
+
+Note that stubs are only written for the parts that we use.
+"""
+
+from typing import Generator, Optional, Union
+
+class Configuration:
+    def ___init__(self, device: Device, configuration: int = 0) -> None: ...
```
kierdavis: There's a lot of definitions in this file that aren't used (as evidenced by the fact that the name of this function contains five underscores and yet it hasn't caused any typechecking errors). I would recommend removing them, so that we're not maintaining things which are not used.

Specifically, I think the following can be removed from this type stub file:

- The `Configuration` class
- `Device.product` and `Device.manufacturer`
- `Device.set_configuration` and `Device.get_active_configuration`
- `Device.reset`

Yes, some of these are useful and we might use them some day. We're not using them right now though, and its easy enough to add solely the required ones back in if/when it turns out we do need them.

stubs/usb/__init__.pyi
```diff
@@ -0,0 +1,8 @@
+"""
+Type stubs for PyUSB.
+
+Note that stubs are only written for the parts that we use.
+"""
+
+import usb.core
+import usb.util
```
kierdavis: Nitpick, but a single trailing newline would be nice here.

stubs/usb/core.pyi
trickeydan: 9d112dc

j5/backends/hardware/sr/v4/power_board.py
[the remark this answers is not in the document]
kierdavis: To clarify - I think passing `0x00` and `0x80` directly to `ctrl_transfer` is the least confusing option here (other options increase cognitive noise without increasing clarity).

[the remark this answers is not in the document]
trickeydan: Will do.

PR #155, opened 2019-03-03
Run LED flashes on wait_start for SR powerboard.

trickeydan:


kierdavis:
Flashing the run LED falls outside the scope of the `wait_until_button_pressed` method of `ButtonInterface`. Somewhat relatedly, `wait_until_button_pressed` doesn't do anything backend-specific - this functionality could easily be implemented at a higher level (e.g. the Board or Robot level).

I'm tempted to say that both of these issues could be solved by binning `ButtonInterface.wait_until_button_pressed` and creating a `PowerBoard.wait_until_start_button_pressed` that does both the polling and the flashing.

kierdavis:
Alternatively, flashing could be isolated from polling by creating a separate thread for it. However, threads are icky (if only we were using `asyncio`!)

trickeydan:
I'd rather keep the `ButtonInterface.wait_until_button_pressed` if we can as I think it could be a useful debug function for competitors.

Perhaps we can keep that **and** create `PowerBoard.wait_until_start_button_pressed` that polls are flashes.

kierdavis:
> Perhaps we can keep that **and** create `PowerBoard.wait_until_start_button_pressed` that polls are flashes.

I'm happy with that.

trickeydan:
Tested working on Hardware

PR #164, opened 2019-03-05
SR v4 Power Board Hardware Tests

trickeydan:


setup.cfg
```diff
@@ -49,5 +50,8 @@ disallow_incomplete_defs = True
 disallow_untyped_decorators = True
 
 check_untyped_defs = True
-
 ignore_missing_imports = False
+
+[mypy-j5.backends.hardware.*]
+# Allows a Callable with Any type params to be used to mock USB.
+disallow_any_generics = False
```
kierdavis: Whitespace error

tests/backends/hardware/sr/v4/test_power_board.py
```diff
@@ -0,0 +1,470 @@
...
+)
+from j5.boards import Board
+from j5.boards.sr.v4.power_board import PowerBoard, PowerOutputPosition
+from j5.components.piezo import Note
+
+
+def test_read_command():
+    """Test that ReadCommand behaves as expected."""
+    rc = ReadCommand(1, 2)
+    assert issubclass(ReadCommand, tuple)
```
kierdavis: I don't forsee us using the `tuple` properties of `ReadCommand` or `WriteCommand`, so I doubt this test is needed. I don't feel strongly about this though.

tests/backends/hardware/sr/v4/test_power_board.py
```diff
@@ -0,0 +1,470 @@
...
+        pass
+
+
+class MockUSBPowerBoardDevice:
+    """This class mocks the behaviour of a USB device for a Power Board."""
+
+    def __init__(self, serial_number: str, fw_version: int = 3):
+        self.serial = serial_number
+        self.firmware_version = fw_version
+        self._ctx = MockUSBContext()
```
kierdavis: `self._ctx` is unused (looks like a remnant from `usb1`). The `MockUSBContext` class can probably be deleted too.

tests/backends/hardware/sr/v4/test_power_board.py
```diff
@@ -0,0 +1,470 @@
...
+        wIndex: int = 0,
+        data_or_wLength: Optional[Union[int, bytes]] = None,
+        timeout: Optional[int] = None,
+    ) -> bytes:
+        """Mock reading data from a device."""
+        assert wValue == 0  # Always 0 on read.
+
+        if 0 <= wIndex < 6:
+            return self.read_output(data_or_wLength)
+        if wIndex == 7:
```
kierdavis: I've noticed a pattern of not using `elif` for mutually exclusive code branches if each block ends in a `return` in a few of your PRs recently. Is this just personal preference or is there a PEP for this that I should be following in my own code?

tests/backends/hardware/sr/v4/test_power_board.py
```diff
@@ -0,0 +1,470 @@
...
+def mock_find(
+    find_all=True, *, idVendor: int, idProduct: int,
+) -> List[MockUSBPowerBoardDevice]:
+    """This function mocks the behaviour of usb.core.find."""
+    assert idVendor == 0x1BDA
+    assert idProduct == 0x0010
+    if find_all:
+        return [MockUSBPowerBoardDevice(f"SERIAL{n}") for n in range(0, 4)]
+    else:
+        return [MockUSBPowerBoardDevice("SERIAL0")]
```
kierdavis: I don't forsee us calling `find(find_all=False)` in the near future, so this `else` block is redundant. I'd suggest just asserting `find_all == True` instead.

Code that never gets executed is doing nothing but waiting to get out of date.

tests/backends/hardware/sr/v4/test_power_board.py
```diff
@@ -0,0 +1,470 @@
...
+
+    assert not backend.get_button_state(MockBoard(), 0)
+
+    with pytest.raises(ValueError):
+        backend.get_button_state(MockBoard(), 1)
+
+
+def test_backend_wait_until_button_pressed():
+    """Test the wait_until_button_pressed."""
+    # No tests here as no known practical method to test it.
```
kierdavis: Maybe we could inject `time.sleep` in the same manner we inject `usb.core.find`, and implement the test like so:

```python
device = MockUSBPowerBoardDevice("SERIAL0")

N = 3
def mock_sleep(secs):
  assert secs <= 0.1 # Check we poll the button at an acceptable rate.
  mock_sleep.times_called += 1
  # Button gets pressed during the N'th call to mock_sleep.
  if mock_sleep.times_called == N:
    device.button_state = True
  # Check mock_sleep doesn't get called more than N times. Prevents wait_until_button_pressed from looping forever.
  assert mock_sleep.times_called <= N
mock_sleep.times_called = 0

backend = SRV4PowerBoardHardwareBackend(device)
backend.wait_until_button_pressed(MockBoard(), 0, sleep=mock_sleep)
assert mock_sleep.times_called == N
```

If the above isn't feasible, please remove this empty test function since it is redundant.

tests/backends/hardware/sr/v4/test_power_board.py
trickeydan: It's used by the USB library in it's cleanup things. An exception is thrown if it isn't there.

kierdavis: In that case please can you add a comment explaining this?

tests/backends/hardware/sr/v4/test_power_board.py
trickeydan: It's just my style. Happy to change if it's disliked.

tests/backends/hardware/sr/v4/test_power_board.py
trickeydan: I attempted this, but it was very dirty. I'm going to accept the lack of coverage.

kierdavis: Yeah, it was a rough idea - I had no idea how well it would pan out in practice.

And yeah, I would prefer us not having coverage of function to us having "fake" coverage (tests that don't verify anything).

PR #169, opened 2019-03-09
Python36

Skybound1:
Works on adding python 3.6 to the supported python versions

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,29 @@
+[tool.poetry]
+name = "j5"
+version = "0.0.2"
+description = "J5 Robotics API"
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -0,0 +1,29 @@
+[tool.poetry]
+name = "j5"
+version = "0.0.2"
+description = "J5 Robotics API"
+authors = ["Dan Trickey <contact@trickey.io>"]
```
trickeydan: We might as well update this to j5 contributors to match the licence

Skybound1: Done

PR #176, opened 2019-03-11
shell.nix: use a more up-to-date URL in workaround command

kierdavis:
@mildlyincompetent could you verify this new command works for you?


mildlyincompetent:
yes it fix

PR #179, opened 2019-03-11
Fix all of the things that Poetry broke.

trickeydan:


trickeydan:
> @Skybound1 FYI you broke this

- @kierdavis

Skybound1:
My bad

PR #193, opened 2019-03-12
Add __str__ to BoardGroup and fix type.

### LINKED ISSUE #72: [FEATURE] BoardGroup __str__
`Board` has `__str__`, so it makes sense for `BoardGroup` to have one too.

trickeydan:
Fixes #72 

Also tightened up a type in BoardGroup constructor that was relevant,


j5/boards/base.py
```diff
@@ -78,8 +78,8 @@ def make_all_safe() -> None:
 class BoardGroup:
     """A group of boards that can be accessed."""
 
-    def __init__(self, board: Board, backend: Backend):
-        self.board_class: Board = board
+    def __init__(self, board: Type[Board], backend: Backend):
```
kierdavis: While we're at it, I think `backend` is supposed to be a `Type[Backend]` here, because we later call `backend.discover()` (or will do after #195 is merged). Also, there isn't a board in this context that `backend` would correspond to if it was an instance rather than a subclass.

j5/boards/base.py
```diff
@@ -104,6 +104,12 @@ def make_safe(self) -> None:
         for board in self.boards.values():
             board.make_safe()
 
+    def __str__(self) -> str:
+        """A string representation of the board group."""
+        list_str = ' , '.join(map(str, self.boards.values()))
```
kierdavis: I feel like the space before the comma here will look weird, but that's just my preference.
```suggestion
        list_str = ', '.join(map(str, self.boards.values()))
```

tests/boards/test_base.py
```diff
@@ -189,6 +189,15 @@ def test_board_group_singular():
     assert type(board_group.singular()) == MockBoard
 
 
+def test_board_group_str():
+    """Test that the board group can be represented as a string."""
+    assert str(BoardGroup(MockBoard, NoBoardMockBackend())) == "Group of Boards - []"
+    assert str(BoardGroup(MockBoard, OneBoardMockBackend())) == \
+        "Group of Boards - [Testing Board - TESTSERIAL1]"
+    assert str(BoardGroup(MockBoard, TwoBoardsMockBackend())) == \
+        "Group of Boards - [Testing Board - TESTSERIAL1 , Testing Board - TESTSERIAL2]"
```
kierdavis: See above.
```suggestion
        "Group of Boards - [Testing Board - TESTSERIAL1, Testing Board - TESTSERIAL2]"
```

j5/boards/base.py
trickeydan: This should be fixed in #195.

trickeydan: I have made #198 to track this.

Issue #116, opened 2019-02-19
Redundant reference to Board is passed to Backend


trickeydan:
e.g
```python
    def __init__(self, identifier: int, board: Board, backend: ServoInterface) -> None:
        self._board = board
        self._backend = backend
        self._identifier = identifier
```

We  then call `self._backend.get_servo_position(self._board, self._identifier)`, but we could access this on `board`.

We would have to fiddle with the types to remove this though.

kierdavis:
Perhaps its the board that's redundant, not the backend? As far as I can recall, boards don't expose methods like `get_servo_position` - they just expose a collection of servo components which talk directly to the implementor of the component's backend interface.

trickeydan:
We should wait until #118 is fixed before continuing on this issue.

trickeydan:
We've discussed this before and I'm going to list the changes that we should make for this:

- `Backend` is now specific to a board instance. Every board instance has it's own Backend instance.
- Thus we don't need to pass `Board` to calls on `Backend` instances as a reference is stored on the `Backend`

Incidentally, this will also fix #153.

kierdavis:
I agree, and I think this is ready to be implemented.

This will also fix #186.

trickeydan:
Fixed by #195

PR #195, opened 2019-03-12
Backend refactor

### LINKED ISSUE #103: Board.discover is redundant
I can't think of any situations where `board.discover(backend)` would do anything except simply call `backend.discover()`, so the former could likely be removed in favour of calling the latter directly.

trickeydan:
Fixes #103, #116


trickeydan:
I've concluded that a `Backend` shouldn't actually need any references to the Board it is controlling. All required data to perform it's actions should be passed in the constructor.

tests/backends/hardware/sr/v4/test_power_board.py
```diff
@@ -354,10 +354,10 @@ def test_backend_get_power_output_enabled():
     backend = SRV4PowerBoardHardwareBackend(device)
 
     for i in range(0, 6):
-        assert not backend.get_power_output_enabled(MockBoard(), i)
```
kierdavis: Please delete the `MockBoard` class, if it is now unused.

trickeydan:
There are a lot of changes in this PR. I'd recommend review commit by commit.

j5/backends/hardware/sr/v4/power_board.py
```diff
@@ -234,56 +233,52 @@ def get_power_output_current(self, board: Board, identifier: int) -> float:
         current, = struct.unpack("<I", self._read(cmd))
         return cast(int, current) / 1000  # convert milliamps to amps
 
-    def buzz(self, board: Board, identifier: int,
-             duration: timedelta, pitch: Pitch) -> None:
+    def buzz(self, identifier: int,
+             duration: timedelta, pitch: int) -> None:
```
kierdavis: If this function now only takes a frequency and no longer accepts a `Note`, please can rename this argument from `pitch` to `frequency`?

j5/boards/board.py
```diff
@@ -78,55 +72,70 @@ def make_all_safe() -> None:
...
+        """The type of board that this group contains."""
+        return self._board_class
+
+    @property
+    def backend_class(self) -> Type[Backend]:
+        """The Backend that this group uses for Boards."""
+        return self._backend_class
+
+    @property
+    def boards(self) -> List[Board]:
```
kierdavis: I think a `Set` would be a more appropriate return type for this function.

trickeydan: Split out into #207

j5/backends/hardware/sr/v4/power_board.py
trickeydan: This change was made in a separate PR, but I'll fix it here anyway

PR #204, opened 2019-03-21
Console environment

### LINKED ISSUE #191: Console Environment / Backend
Related to #108 

This will be really useful for dev and debugging

trickeydan:
Do not merge until #195 has been merged.

Fixes #191


j5/backends/console/env.py
```diff
@@ -0,0 +1,42 @@
...
+
+from j5.backends import Environment
+
+ConsoleEnvironment = Environment("ConsoleEnvironment")
+
+T = TypeVar("T")
+
+
+class Console:
+    """A helper class for the comsole environment."""
```
kierdavis: Typo
```suggestion
    """A helper class for the console environment."""
```

j5/backends/console/env.py
```diff
@@ -0,0 +1,42 @@
...
+    def info(self, message: str) -> None:
+        """Print information to the user."""
+        self._print(f"{self._descriptor}: {message}")
+
+    def read(self, prompt: str, return_type: Type[T] = str) -> T:  # type: ignore
+        """
+        Get a value of type 'return_type'.
+
+        We have to ignore the types on this function unfortunately, as static type
+        checking is not powerful enough to confirm that it is correct at runtime.
```
kierdavis: Nitpick: I think remarks about the implementation details belong in comments, rather than in the docstring.

j5/backends/console/sr/v4/power_board.py
```diff
@@ -0,0 +1,146 @@
...
+
+class SRV4PowerBoardConsoleBackend(
+    PowerOutputInterface,
+    PiezoInterface,
+    ButtonInterface,
+    BatterySensorInterface,
+    LEDInterface,
+    Backend,
+):
+    """The hardware implementation of the SR V4 power board."""
```
kierdavis: s/hardware/console/
```suggestion
    """The console implementation of the SR V4 power board."""
```

j5/backends/console/sr/v4/power_board.py
```diff
@@ -0,0 +1,146 @@
...
+        except KeyError:
+            raise ValueError(f"Invalid power output identifier {identifier!r}; "
+                             f"valid identifiers are "
+                             f"{self._output_states.keys()}") from None
+
+    def get_power_output_current(self, identifier: int) -> float:
+        """Get the current being drawn on a power output, in amperes."""
+        if identifier in self._output_states:
+
+            return self._console.read(f"Power Output current for {identifier}")
```
kierdavis: - I think that this would be better wording given that `identifier` is an integer rather than an output name.
- I think we should make it clear to the user what they're expected to type in (i.e. type or units of value).
- This should be returning a `float` not a `str`.

```suggestion
            return self._console.read(f"Current for power output {identifier} [amps]", float)
```

j5/backends/console/sr/v4/power_board.py
```diff
@@ -0,0 +1,146 @@
...
+        if duration_ms > 65535:
+            raise ValueError("Maximum piezo duration is 65535ms.")
+        self._console.info(f"Buzzing at {pitch}Hz for {duration_ms}ms")
+
+    def get_button_state(self, identifier: int) -> bool:
+        """Get the state of a button."""
+        if identifier != 0:
+            raise ValueError(f"invalid button identifier {identifier!r}; "
+                             f"the only valid identifier is 0")
+        return self._console.read("Start button state", bool)
```
kierdavis: I think we should make it clear to the user what they're expected to type in (i.e. type or units of value).
```suggestion
        return self._console.read("Start button state [true/false]", bool)
```

j5/backends/console/sr/v4/power_board.py
```diff
@@ -0,0 +1,146 @@
...
+
+    def wait_until_button_pressed(self, identifier: int) -> None:
+        """Halt the program until this button is pushed."""
+        self._console.info("Waiting for start button press.")
+        while not self.get_button_state(identifier):
```
kierdavis: Maybe it would be more user-friendly to do something like
```python
self._console.read("Hit return to press start button")
```
instead of looping here.

j5/backends/console/sr/v4/power_board.py
```diff
@@ -0,0 +1,146 @@
...
+            sleep(0.05)
+
+    def get_battery_sensor_voltage(self, identifier: int) -> float:
+        """Get the voltage of a battery sensor."""
+        if identifier != 0:
+            raise ValueError(f"invalid battery sensor identifier {identifier!r}; "
+                             f"the only valid identifier is 0")
+        return self._console.read("Battery Sensor Voltage", float)
```
kierdavis: I think we should make it clear to the user what they're expected to type in (i.e. type or units of value).
```suggestion
        return self._console.read("Battery voltage [volts]", float)
```

j5/backends/console/sr/v4/power_board.py
```diff
@@ -0,0 +1,146 @@
...
+
+    def get_battery_sensor_current(self, identifier: int) -> float:
+        """Get the current of a battery sensor."""
+        if identifier != 0:
+            raise ValueError(f"invalid battery sensor identifier {identifier!r}; "
+                             f"the only valid identifier is 0")
+        return self._console.read("Battery Sensor Current", float)
```
kierdavis: I think we should make it clear to the user what they're expected to type in (i.e. type or units of value).
```suggestion
        return self._console.read("Battery current [amps]", float)
```

tests/backends/console/test_env.py
```diff
@@ -0,0 +1,45 @@
+"""Tests for the ConsoleEnvironment and Console helper."""
+
+from j5.backends import Environment
+from j5.backends.console import Console, ConsoleEnvironment
+
+
+def test_hardware_environment():
```
kierdavis: s/hardware/console/
```suggestion
def test_console_environment():
```

tests/backends/console/test_env.py
```diff
@@ -0,0 +1,45 @@
...
+    """Test that the Hardware Environment works."""
```
kierdavis: s/hardware/console/
```suggestion
    """Test that the Console Environment works."""
```

tests/backends/console/sr/__init__.py
```diff
@@ -0,0 +1 @@
+"""Test SR sonsole backends."""
```
kierdavis: Typo
```suggestion
"""Test SR console backends."""
```

j5/backends/console/sr/v4/power_board.py
```diff
@@ -78,8 +77,9 @@ def set_power_output_enabled(
         """Set whether a power output is enabled."""
         try:
             self._console.info(f"Setting output {identifier} to {enabled}")
+            assert identifier in self._output_states.keys()
             self._output_states[identifier] = enabled
-        except KeyError:
+        except AssertionError:
```
kierdavis: Any reason for this change? I don't have anything against it really, I'm just interested to know.

trickeydan: Due to the implementation of the `_output_states` on this backend, Python will not throw a `KeyError` when `self._output_states[identifier] = enabled` is called.

In order to test this properly, I added `assert identifier in self._output_states.keys()`, which is then caught when it fails.

kierdavis: Ah that makes total sense, thanks!

j5/backends/console/sr/v4/power_board.py
```diff
@@ -0,0 +1,148 @@
...
+                             f"valid identifiers are "
+                             f"{self._output_states.keys()}") from None
+
+    def set_power_output_enabled(
+        self, identifier: int, enabled: bool,
+    ) -> None:
+        """Set whether a power output is enabled."""
+        try:
+            self._console.info(f"Setting output {identifier} to {enabled}")
+            assert identifier in self._output_states.keys()
```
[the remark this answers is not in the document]
trickeydan: Ah I wasn't aware of this. I'll replace with an if statement.

j5/backends/console/sr/v4/power_board.py
```diff
@@ -0,0 +1,146 @@
...
+            for i in range(2)
+        }
+
+        # Setup console helper
+        self._console = console_class(f"{self.board.__name__}({self._serial})")
+
+    @property
+    def firmware_version(self) -> Optional[str]:
+        """The firmware version reported by the board."""
+        return None  # Console, so no firmware
```
sedders123: Could we return the j5 version? Not sure if useful but might make debugging which version is in use easier?

tests/backends/console/sr/v4/test_power_board.py
```diff
@@ -0,0 +1,229 @@
...
+from j5.components.piezo import Note
+
+T = TypeVar("T")
+
+
+class MockConsole(Console):
+    """Test Console for testing the board."""
+
+    def __init__(self, descriptor: str):
+        super(Console, self).__init__()
```
sedders123: As we're using Python 3 I believe we can just use `super().__init__()`

PR #212, opened 2019-03-31
Motors

### LINKED ISSUE #41: Create MotorOutput Component

trickeydan:
Fixes #41


j5/components/motor_output.py
```diff
@@ -0,0 +1,75 @@
...
+        """Get if the motor is braked."""
+        pass
+
+    @abstractmethod
+    def set_motor_output_coast(self, board: Board, identifier: int):
+        """Coast the motor output."""
+        pass
+
+    @abstractmethod
+    def get_motor_output_coast(self, board: Board, identifier: int):
```
kierdavis: I'm not sure how I feel about splitting the different states of a motor ("on", "brake", "coast") into different methods like this. In particular, implementing `MotorOutput.power` would require calling at least two (if not all three) of the `get_*` methods in order to work out what state the motor is in. Is there a reason not to use `MotorOutputState` here?

j5/components/motor_output.py
```diff
@@ -0,0 +1,75 @@
...
+        """Get if the motor is coasting."""
+        pass
+
+
+class MotorOutput(Component):
```
kierdavis: I'd suggest simply calling this `Motor` (and renaming the interface and `MotorOutputState` stuff appropriately), to be consistent with the [`Servo` component](https://github.com/j5api/j5/blob/master/j5/components/servo.py#L31).

j5/components/motor_output.py
trickeydan: 4a31bc0

kierdavis:
Looks good so far :+1:

tests/components/test_motor.py
```diff
@@ -0,0 +1,128 @@
...
+
+    assert motor0.state == MotorSpecialState.COAST
+    assert motor1.state == MotorSpecialState.BRAKE
+
+
+def test_motor_set_state() -> None:
+    """Test that we can set the state of a motor."""
+    motor = Motor(0, MockMotorBoard(), MockMotorDriver())
+
+    motor.state = 0
```
kierdavis: I'd recommend also testing that the limits (+1 and -1) and fractional values are valid.
```suggestion
    motor.state = 0
    motor.state = 1
    motor.state = -1
    motor.state = 0.123
```

tests/components/test_motor.py
```diff
@@ -0,0 +1,128 @@
...
+    motor1 = Motor(3, MockMotorBoard(), MockMotorDriver())
+
+    assert type(motor0.state) is MotorSpecialState
+    assert type(motor1.state) is MotorSpecialState
+
+    assert motor0.state == MotorSpecialState.COAST
+    assert motor1.state == MotorSpecialState.BRAKE
+
+
+def test_motor_set_state() -> None:
```
kierdavis: It irks me that there's no assertions in this test (yeah, I know its really just testing that no exceptions are being raised). Perhaps we could make `MockMotorDriver` save the last state written to it, and check in this function that it is equal to what we expect it to be?

e.g.

```python
driver = MockMotorDriver()
motor = Motor(0, MockMotorBoard(), driver)
...
motor.state = 0.123
assert driver.last_written_state == 0.123
```

trickeydan: I suggest making a separate issue for this, as I don't think any of the other tests behave this way?

kierdavis: Huh, you are correct. I think this is something we really should fix.

Opened #216.

j5/components/motor.py
```diff
@@ -0,0 +1,64 @@
...
+    def state(self) -> MotorState:
+        """Get the current state of this output."""
+        return self._backend.get_motor_state(self._identifier)
+
+    @state.setter
+    def state(self, new_state: MotorState) -> None:
+        """Set the current state of this output."""
+        if isinstance(new_state, float):
+            if new_state < -1 or new_state > 1:
+                raise ValueError("Motor state must be between 1 and -1.")
```
kierdavis: A motor state can also be `BRAKE` or `COAST`, which are not between 1 and -1.

```suggestion
                raise ValueError("Motor speed must be between 1 and -1.")
```

PR #220, opened 2019-04-08
Better development docs

trickeydan:


trickeydan:
Merged manually without approval, as this is just docs.

docs/development/getting-started.rst
```diff
@@ -0,0 +1,82 @@
...
+-------------------
+
+`j5` is developed on GitHub_ and pull requests should be submitted there. If you have write access to the repository,
+you optionally can develop your changes on a branch within the main repository. Alternatively, please fork the `j5`
+repository and pull request from there.
+
+If you are working on something that has an existing issue open on the `j5` repository, please ensure that you assign
+the issue to yourself such that duplication of work does not accidentally occur.
+
+If you need help with Git, there are plenty of good tutorials online.
```
kierdavis: I think it would be more approachable for newcomers if we recommended one or two tutorials here.

I've opened #228 to track this.

docs/development/getting-started.rst
```diff
@@ -0,0 +1,82 @@
...
+Setting Up
+----------
+
+Development of `j5` has the same basic requirements as usage of the API.
+
+You will need the following installed on your machine:
+
+- Python_ 3.7 or higher
+- python3-pip (for package management)
+- pipenv_ (optional)
```
kierdavis: We use `poetry` instead of `pipenv` now.

I've opened #229 to track this.

PR #225, opened 2019-04-09
Correct version and copyright for docs

trickeydan:


j5/__init__.py
```diff
@@ -5,5 +5,5 @@
 
 __all__ = ["BoardGroup", "BaseRobot", "__version__", "__version_short__"]
 
-__version__ = "0.0.2"
-__version_short__ = "0.0.2"
+__version__ = "0.1.0"
```
sedders123: Could we use `j5.__version__` and `j5.__version_short__` here to minimise the number of places we have to remember when releasing a new version?

PR #219, opened 2019-04-08
Minor docs improvements

trickeydan:
Fixes #187, #188, #189

Issue #194, opened 2019-03-12
Document Supported Boards / Hardware


trickeydan:
Relevant to #139

PR #226, opened 2019-04-09
Document our supported hardware

### LINKED ISSUE #139: Document how to udev allow PowerBoard, etc

trickeydan:
Fixes #139, #194 

Do not merge before #219, as this was branched from that at 62a39ae


trickeydan:
Merging without approval as this is just docs, mostly verbatim from SR.

trickeydan:
> I'm not sure whether we need to duplicate the power board details when we could just link to the SR website instead.

I'm mostly concerned about the SR docs disappearing, as old documentation (e.g v3) has been known to in the past. I also intend to develop this section into a much more technical overview of the boards in the future, not aimed at competitors like the SR docs are.

PR #227, opened 2019-04-10
SR v4 Motor Board

### LINKED ISSUE #63: Support SR v4 Motor Board
Requires MotorOutput - #41

trickeydan:
- [x] Motor Board Class
- [x] Motor Board Class tests
- [x] Hardware Implementation
- [x] Hardware Implementation tests
- [x] Console Implementation
- [x] Console Implementation tests

This is a huge PR, commit - by - commit review is recommended. Putting docs separately as this is already big.

Closes #63


j5/backends/console/sr/v4/motor_board.py
```diff
@@ -0,0 +1,69 @@
...
+        return self._state[identifier]
+
+    def set_motor_state(self, identifier: int, power: MotorState) -> None:
+        """Set the state of a motor."""
+        if identifier not in range(0, 2):
+            raise ValueError(
+                f"Invalid motor identifier: {identifier}, valid values are: 0, 1",
+            )
+        self._state[identifier] = power
+        if isinstance(power, MotorSpecialState):
```
Skybound1: `try: except:` ?

j5/backends/hardware/sr/v4/motor_board.py
```diff
@@ -0,0 +1,190 @@
...
+    This is a decorator that should be used to wrap any functions that call the serial
+    interface. It will catch and rethrow the errors as a CommunicationError, so that it
+    is more explicit what is going wrong.
+    """
+    @wraps(func)
+    def catch_exceptions(*args, **kwargs):  # type: ignore
+        try:
+            return func(*args, **kwargs)
+        except SerialTimeoutException as e:
+            raise CommunicationError(f"Serial Timeout Error: {e}")
```
Skybound1: Is it worth raising these `from None`?

j5/backends/hardware/sr/v4/motor_board.py
```diff
@@ -0,0 +1,190 @@
...
+            cls,
+            find: Callable = comports,
+            serial_class: Type[Serial] = Serial,
+    ) -> List[Board]:
+        """Discover all connected motor boards."""
+        # Find all serial ports.
+        ports: List[ListPortInfo] = find()
+
+        # Filter down to just motor boards.
+        ports = list(filter(lambda x: x.vid == 0x403, ports))  # FTDI, Ltd
```
Skybound1: Instead of having a ton of `ports = list(filter(`, you could have a function defined here that does all the checks, and then have a single filter

Skybound1: Or you could have a dictionary that states what each of the attrs needs to be, then iterate over that with an `all`

j5/backends/hardware/sr/v4/motor_board.py
```diff
@@ -0,0 +1,190 @@
...
+        if len(bdata) == 0:
+            raise CommunicationError(
+                "Unable to communicate with motor board. ",
+                "Is it correctly powered?",
+            )
+
+        ldata = bdata.decode('utf-8')
+        return ldata.rstrip()
+
+    def get_firmware_version(self) -> Optional[str]:
```
Skybound1: AFAIK in python we don't typically _aim_ for having `get_` or `set_`, instead using `@property` here with just `firmware_version` might be more pythonic?

j5/backends/hardware/sr/v4/motor_board.py
```diff
@@ -0,0 +1,190 @@
...
+            port=serial_port,
+            baudrate=1000000,
+            timeout=0.25,
+        )
+
+        # Check we have the correct firmware version.
+        version = self.get_firmware_version()
+        if version != "3":
+            raise CommunicationError(
+                f"Unexpected firmware version: {version}, expected: 3.",
```
Skybound1: The comparison above is for a string, this message suggests int?

j5/backends/hardware/sr/v4/motor_board.py
```diff
@@ -0,0 +1,190 @@
...
+            )
+        return firmware_data[6:]  # Strip model and return version
+
+    def get_motor_state(self, identifier: int) -> MotorState:
+        """Get the state of a motor."""
+        # We are unable to read the state from the motor board,
+        # so we'll get the last set value.
+        return self._state[identifier]
+
+    def set_motor_state(self, identifier: int, power: MotorState) -> None:
```
Skybound1: Something to consider, instead of `set_motor_state`, what if this class actually exposed itself as a list of `self._state`, ie if you instantiatied this class and called it `foo`. `foo[0].motor_state = power`. and then the `motor_state` can be a property setter of the `MotorState` class.. or something along those lines at least?

j5/boards/sr/v4/motor_board.py
```diff
@@ -0,0 +1,55 @@
...
+
+    @property
+    def serial(self) -> str:
+        """Get the serial number."""
+        return self._serial
+
+    @property
+    def firmware_version(self) -> Optional[str]:
+        """Get the firmware version of the board."""
+        return self._backend.get_firmware_version()
```
Skybound1: I'd recommend trying to get this is a property, like the thing it's in ;)

j5/backends/hardware/sr/v4/motor_board.py
sedders123: I'd argue the opposite. If one of the aims of j5 is to empower competitors to explore their kit in depth, hiding error messages from them seems counter intuitive. Also, even if teams are unable technically to dig into an issue this seems like useful information for them to forward on to us if they encounter a weird issue.

I reckon that most of our wrapped errors should be `from e` so that teams get as much information as we can give them about why their robot crashed

trickeydan: I have to agree with @sedders123 here. I think that we should aim to expose as much information to competitors as possible to help them debug their problems. The reason for raising a `CommunicationError` here is so that the amount of information does not become overwhelming, there is a simple message at the end of the error stating that the serial interface timed out.

There is scope for a debug mode that raises `from e` and otherwise `from None`, but I think that's a separate issue.

j5/backends/hardware/sr/v4/motor_board.py
trickeydan: So we decided a while back that a `firmware_version` is a `str`, would changing the error to `f"Unexpected firmware version: {version}, expected: \"3\"."` help?

j5/backends/hardware/sr/v4/motor_board.py
trickeydan: So this is part of a separate issue that bugged me whilst writing this. I have written the code here to match the existing style and opened #236 to track the overall issue.

j5/backends/hardware/sr/v4/motor_board.py
trickeydan: I personally like the concept as it is more pythonic, although I am unsure how it would work in this instance.

Remember, a `Backend` is never directly called by a competitor. It is an abstract interface that is called by the relevant `Component`. In this case, the `set_motor_state` function is called by the `Motor.state` setter.

j5/boards/sr/v4/motor_board.py
trickeydan: Opened #236. Leaving this as is to match existing code style.

j5/backends/hardware/sr/v4/motor_board.py
kierdavis: I wrote a bunch of stuff about this before I saw @Skybound1's comment, so here's the interesting bit:

```python
def is_motor_board(port):
  return port.manufacturer == "Student Robotics" and port.product == "MCV4B" \
    and port.vid == 0x0403 and port.pid == 0x6001
...
for port in filter(is_motor_board, comports()):
  ...
```

Using `all` and a dictionary is a nice approach, though I personally think its overly complex for this use.

j5/boards/sr/v4/motor_board.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def make_safe(self) -> None:
+        """Make this board safe."""
+        for output in self._outputs.values():
+            # Brake both motors.
+            output.state = MotorSpecialState.BRAKE
```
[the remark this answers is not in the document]
trickeydan: From `robotd`:
```python
# set both motors to brake
self.connection.write(b'\x00\x02\x02\x03\x02')
self._status = {'m0': 'brake', 'm1': 'brake'}
```
This was authored by Alistair Lynn, and I generally trust his approach.

I do however see the merits of using `COAST` instead.

If we consider a crashed robot in the arena, do we want it to stop as quickly as possible? or come to a slow stop?

For context, `sr.robot` has no safety features like this.

j5/backends/hardware/sr/v4/motor_board.py
Skybound1: I was thinking something along the lines of 

```
parameters = {
    'manufacturer': 'Student Robotics',
    'product': 'MCV48,
...
}
for port in comports:
    if all(getattr(port, k) == v for k, v in parameters.keys():
```

you this is getting too crazy xD

j5/backends/hardware/sr/v4/motor_board.py
Skybound1: It would, would the user be able to manipulate firmware_version by any chance? Wondering if there is any point to this

trickeydan: `firmware_version` should be a read-only attribute

j5/backends/hardware/sr/v4/motor_board.py
```diff
@@ -0,0 +1,190 @@
...
+        """Send a serial command to the board."""
+        command_bytes = chr(command).encode('utf-8')
+        bytes_written = self._serial.write(command_bytes)
+        if len(command_bytes) != bytes_written:
+            raise CommunicationError(
+                "Mismatch in command bytes written to serial interface.",
+            )
+
+        if data is not None:
+            data_bytes = chr(data).encode()
```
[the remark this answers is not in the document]
trickeydan: This is much nicer.

j5/backends/hardware/sr/v4/motor_board.py
trickeydan: For this pull request, I'll go with @kierdavis's approach, but if we refactor to have a generic serial device backend in the future, I can see the merits of @Skybound1's approach.

j5/backends/console/sr/v4/motor_board.py
trickeydan: How would you do that without breaking the typing?

The following doesn't pass `mypy`.

```python
        try:
            power_human_name = power.name
        except AttributeError:
            power_human_name = str(power)
```

Skybound1: What's the error?

trickeydan: `j5/backends/console/sr/v4/motor_board.py:66: error: Item "float" of "Union[float, MotorSpecialState]" has no attribute "name"`

j5/backends/hardware/sr/v4/motor_board.py
trickeydan: 46c3740

j5/backends/hardware/sr/v4/motor_board.py
trickeydan: 4b2cfd1

j5/backends/console/sr/v4/motor_board.py
Skybound1: Hmm... this feels like an issue here with `mypy` as it isn't recognising the try / except, except I don't think I would expect it to... leading to is it too much to permit both floats and MotorSpecialState... ? 

As I'm lazy, I think we should just go with look before you leap for this case save further effort from discussion?

j5/backends/hardware/sr/v4/motor_board.py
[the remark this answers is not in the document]
trickeydan: 21ee864

j5/backends/hardware/sr/v4/motor_board.py
```diff
@@ -0,0 +1,190 @@
...
+        for i, val in enumerate(self._state):
+            self.set_motor_state(i, MotorSpecialState.BRAKE)
+
+        self._serial.flush()
+        self._serial.close()
+
+    @handle_serial_error
+    def send_command(self, command: int, data: Optional[int] = None) -> None:
+        """Send a serial command to the board."""
+        command_bytes = chr(command).encode('utf-8')
```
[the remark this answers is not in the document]
trickeydan: 21ee864

j5/backends/console/sr/v4/motor_board.py
trickeydan: See https://github.com/python/mypy/issues/2420

Using `if isinstance ...` appears to be the correct approach according to BDFL.

j5/boards/sr/v4/motor_board.py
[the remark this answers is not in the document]
kierdavis: I think resolving this is outside the scope of this PR, and possibly even outside the scope of j5 (i.e. we should leave it up to the competition vendor to decide).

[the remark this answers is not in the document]
trickeydan: In that case, should we open an issue for a safe_mode on the Motor Component and merge this?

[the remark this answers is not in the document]
kierdavis: Created #250 to track this.

kierdavis:
Looks like some things broke when #233 (change lists to sets) was merged into this branch. Once these are fixed I think this PR is good to go.

PR #233, opened 2019-04-13
Use sets instead of lists for unordered collections

### LINKED ISSUE #207: Use set rather than list for unordered collections
I think a `Set` would be a more appropriate return type for this function.

_Originally posted by @kierdavis in https://github.com/j5api/j5/pull/195_

kierdavis:
Our code uses lists in a lot of places where a set would be more appropriate, due to the collection not having a useful order to it. This PR changes the following methods/properties to be sets instead of lists: `Board.supported_components`, `Backend.discover`, `Environment.supported_boards` and `Board.BOARDS`. Resolves #207.

I'd recommend reviewing this PR commit-by-commit.

Note: I'd also like to change `supported_modes` in the `GPIOPin` component, but I don't know what to do about the fact that `initial_mode` defaults to the first item of `supported_modes`, implying that `supported_modes` is actually ordered. Possible solutions may be to require `initial_mode` to be explicitly given, or to determine its default value some other way, or to leave this bit of code as it is. Thoughts?

### DIFF
--- a/tests/backends/hardware/sr/v4/test_power_board.py
+++ b/tests/backends/hardware/sr/v4/test_power_board.py
@@ -277,7 +277,7 @@ def test_backend_discover() -> None:
     found_boards = SRV4PowerBoardHardwareBackend.discover(find=mock_find)
 
     assert len(found_boards) == 4
-    assert type(found_boards[0]) is PowerBoard
+    assert all(type(board) is PowerBoard for board in found_boards)
 
 
 def test_backend_cleanup() -> None:
--- a/j5/backends/backend.py
+++ b/j5/backends/backend.py
@@ -1,7 +1,7 @@
 """The base classes for backends."""
 
 from abc import ABCMeta, abstractmethod
-from typing import TYPE_CHECKING, Dict, List, Optional, Set, Type
+from typing import TYPE_CHECKING, Dict, Optional, Set, Type
 
 if TYPE_CHECKING:
     from j5.boards import Board  # noqa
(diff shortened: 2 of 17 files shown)


Skybound1:
I guess my question would be in what way is set better for this particular use case? I get that we don't need an order. Are we needing to do checks where we check if an object is in a list / set? If it's just iterating, I believe lists are a tad quicker, so curious to know what the perceived gain would be

trickeydan:
> I guess my question would be in what way is set better for this particular use case? I get that we don't need an order. Are we needing to do checks where we check if an object is in a list / set? If it's just iterating, I believe lists are a tad quicker, so curious to know what the perceived gain would be

I believe we do check if things are in the sets in this instance, although we do also iterate over them.

According to the [Python Wiki](https://wiki.python.org/moin/TimeComplexity), sets are implemented using a [hash table](https://en.wikipedia.org/wiki/Hash_table), and are very similar to the implementation for `dict`.

Even if `set` were to be slower than `list` in these cases, I personally feel that it is better to use a set, so that students / users do not perceive that there is an order when there is not.

tests/backends/hardware/sr/v4/test_power_board.py
```diff
@@ -277,7 +277,7 @@ def test_backend_discover() -> None:
     found_boards = SRV4PowerBoardHardwareBackend.discover(find=mock_find)
 
     assert len(found_boards) == 4
-    assert type(found_boards[0]) is PowerBoard
+    assert all(type(board) is PowerBoard for board in found_boards)
```
trickeydan: This is much nicer.

trickeydan:
> Note: I'd also like to change `supported_modes` in the `GPIOPin` component, but I don't know what to do about the fact that `initial_mode` defaults to the first item of `supported_modes`, implying that `supported_modes` is actually ordered. Possible solutions may be to require `initial_mode` to be explicitly given, or to determine its default value some other way, or to leave this bit of code as it is. Thoughts?

I think that it's probably worth a refactor of this section of the codebase to use a set here, as I agree that the `supported_modes` should not be ordered. I'd be in favour of explicitly stating the `initial_mode`.

kierdavis:
> I think that it's probably worth a refactor of this section of the codebase to use a set here, as I agree that the supported_modes should not be ordered. I'd be in favour of explicitly stating the initial_mode.

Tracked in #237.

Issue #216, opened 2019-04-03
Ensure tests for setters/actions check desired result is achieved


kierdavis:
Many of our tests run some code but don't use any assertions to check the code behaved as expected. Essentially these tests are only checking that no exceptions are raised.

This is common among, but not limited to, tests for performing actions on an object or changing its state - we should be checking that the action had the desired effect.

Examples:

* [`test_servo_set_position`](https://github.com/j5api/j5/blob/2d9c1faf4910f6c3004b48e6da93ba8b884badc6/tests/components/test_servo.py#L82)
* [`test_motor_set_state`](https://github.com/j5api/j5/blob/49c2f602e9bc82c80c06f912edefb2fd028c409b/tests/components/test_motor.py#L110)
* [`test_backend_piezo_buzz`](https://github.com/j5api/j5/blob/2d9c1faf4910f6c3004b48e6da93ba8b884badc6/tests/backends/hardware/sr/v4/test_power_board.py#L350)
* [`test_power_board_make_safe`](https://github.com/j5api/j5/blob/2d9c1faf4910f6c3004b48e6da93ba8b884badc6/tests/boards/sr/v4/test_power_board.py#L119) (this doesn't check that the outputs are switched off, which is the whole point of `make_safe`! :scream:)

trickeydan:
Relevent: #214

PR #235, opened 2019-04-14
Improve test coverage of various things

trickeydan:
Recommend to review commit-by-commit, as some things are completely unrelated.

### DIFF
--- a/tests/backends/console/test_env.py
+++ b/tests/backends/console/test_env.py
@@ -43,3 +43,14 @@ def test_console_read() -> None:
     console = Console("TestBoard", input_function=mock_input)
 
     assert str(console.read("Enter Test Input")) == str(reversed("Enter Test Input"))
+
+
+def test_console_read_bad_type() -> None:
+    """Test that a ValueError is thrown if we can't construct the right type."""
+    # Define a testing input function
+    def mock_input(prompt: str) -> str:
+        """Mock some input."""
+        return "string"
+
+    console = Console("TestBoard", input_function=mock_input)
+    assert console.read("Enter test input", None) is None
--- a/tests/backends/console/test_env.py
+++ b/tests/backends/console/test_env.py
@@ -45,8 +45,8 @@ def test_console_read() -> None:
     assert str(console.read("Enter Test Input")) == str(reversed("Enter Test Input"))
 
 
-def test_console_read_bad_type() -> None:
-    """Test that a ValueError is thrown if we can't construct the right type."""
+def test_console_read_none_type() -> None:
+    """Test that we can read None from console, i.e any input."""
     # Define a testing input function
     def mock_input(prompt: str) -> str:
         """Mock some input."""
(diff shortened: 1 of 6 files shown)


tests/backends/console/test_env.py
```diff
@@ -43,3 +43,14 @@ def mock_input(prompt: str) -> str:
     console = Console("TestBoard", input_function=mock_input)
 
     assert str(console.read("Enter Test Input")) == str(reversed("Enter Test Input"))
+
+
+def test_console_read_bad_type() -> None:
+    """Test that a ValueError is thrown if we can't construct the right type."""
```
kierdavis: This docstring doesn't seem to match the rest of the function.

tests/boards/sr/v4/test_power_board.py
```diff
@@ -174,3 +174,11 @@ def test_power_board_error_led() -> None:
     pb = PowerBoard("SERIAL0", MockPowerBoardBackend())
 
     assert type(pb._error_led) is LED
+
+
+def test_power_board_wait_start() -> None:
+    """Test the wait_for_start_flash method."""
+    pb = PowerBoard("SERIAL0", MockPowerBoardBackend())
+
+    # Note: This isn't a great test, but ensures that the code runs at least.
```
kierdavis: Relevant: #216.

tests/components/test_power_output.py
```diff
@@ -88,3 +92,35 @@ def test_power_output_current() -> None:
     power_output = PowerOutput(0, MockPowerOutputBoard(), MockPowerOutputDriver())
     assert type(power_output.current) is float
     assert power_output.current == 8.1
+
+
+def test_power_output_group_instantiation() -> None:
+    """Test that we can instantiate a PowerOutput group."""
+    board = MockPowerOutputBoard()
+    outputs = {i: PowerOutput(i, board, MockPowerOutputDriver()) for i in range(0, 5)}
```
kierdavis: Considering that this PR also changes `MockPowerOutputDriver` to make it behave independently when different `identifier`s are given, I think that this line should use the same instance of `MockPowerOutputDriver` for all outputs rather than creating new ones each time.

```suggestion
    backend = MockPowerOutputDriver(5)
    outputs = {i: PowerOutput(i, board, backend) for i in range(0, 5)}
```
and similarly for the other two test cases.

tests/backends/test_backend.py
```diff
@@ -117,3 +118,97 @@ def test_environment_board_get_backend_unknown() -> None:
...
+                return None
+
+
+def test_backend_check_multiple_backends_same_env() -> None:
+    """Test that we can't define two backends for the same board/environment combo."""
+    test_environment = Environment("test_environment")
+
+    with pytest.raises(RuntimeError):
+        class BackendOne(Backend):
+            """I have no environment."""
```
kierdavis: This docstring doesn't match the rest of the class (but I don't think it needs a docstring anyway).

kierdavis:
I like this - better tests are always good! There's a few fixes to be made though.

tests/backends/console/test_env.py
trickeydan: 052df9a

PR #243, opened 2019-04-16
Use floating point numbers for Piezo

### LINKED ISSUE #125: Piezo frequencies could be floats
Hypothetically, piezo implementations may support more precision than an integer allows, and the frequencies of musical notes we provide in the `Note` enum could be specified to more precision than they currently are. It seems reasonable to use floats everywhere and make it the responsibility of the backend to round to an integer if it does not support fractional frequencies.

This should probably be implemented before or at the same time that #122 is, to avoid having to change things twice.

trickeydan:
Resolves #125

### DIFF
--- a/j5/backends/hardware/sr/v4/power_board.py
+++ b/j5/backends/hardware/sr/v4/power_board.py
@@ -242,7 +242,8 @@ class SRV4PowerBoardHardwareBackend(
         duration_ms = round(duration / timedelta(milliseconds=1))
         if duration_ms > 65535:
             raise ValueError("Maximum piezo duration is 65535ms.")
-        data = struct.pack("<HH", int(frequency), duration_ms)
+        frequency_int = int(round(frequency))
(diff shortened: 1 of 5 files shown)


j5/backends/hardware/sr/v4/power_board.py
```diff
@@ -234,15 +234,15 @@ def get_power_output_current(self, identifier: int) -> float:
...
+             duration: timedelta, frequency: float) -> None:
         """Queue a pitch to be played."""
         if identifier != 0:
             raise ValueError(f"invalid piezo identifier {identifier!r}; "
                              f"the only valid identifier is 0")
         duration_ms = round(duration / timedelta(milliseconds=1))
         if duration_ms > 65535:
             raise ValueError("Maximum piezo duration is 65535ms.")
-        data = struct.pack("<HH", frequency, duration_ms)
+        data = struct.pack("<HH", int(frequency), duration_ms)
```
kierdavis: Nitpick: could we round instead of flooring?

PR #251, opened 2019-04-19
Remove redundant reference to board from components

### LINKED ISSUE #215: Unnecessary reference to Board on Components

kierdavis:
Fixes #215.

### DIFF
--- a/j5/components/gpio_pin.py
+++ b/j5/components/gpio_pin.py
@@ -4,7 +4,6 @@ from abc import abstractmethod
 from enum import IntEnum
 from typing import List, Optional, Type
 
-from j5.boards import Board
 from j5.components.component import (
     Component,
     Interface,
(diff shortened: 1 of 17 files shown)


j5/components/gpio_pin.py
```diff
@@ -136,8 +133,7 @@ def mode(self, pin_mode: GPIOPinMode) -> None:
         """Set the hardware mode of this pin."""
         if pin_mode not in self._supported_modes:
             raise NotSupportedByHardwareError(
-                f"Pin {self._identifier} on {str(self._board)} \
-                does not support {str(pin_mode)}.",
+                f"Pin {self._identifier} does not support {str(pin_mode)}.",
```
trickeydan: I think that losing the board that this pin is on is a slight loss, but that this also opens up a `Component` to be independent of a `Board`, which I approve of.

PR #254, opened 2019-04-19
TEST PR DO NOT MERGE

kierdavis:
This PR adds a new board that still uses a list for its
`supported_components`. I expect this to fail CI because since #233,
`supported_components` needs to be a set.


kierdavis:
Turns out this wasn't actually needed for testing #257.

PR #256, opened 2019-04-19
TEST PR DO NOT MERGE

kierdavis:
This PR contains the same content as #254; however it is based
off an old revision of master from before #233 was merged.
Right now I expect this to pass CI because this new board agrees
with the rest of the codebase in this branch. However, merging
this with master pulls in #233 and hence causes CI to fail. We
need to catch this before the PR gets merged.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,6 +12,13 @@ jobs:
       - image: j5api/ci:3.6
     steps:
       - checkout
+      - run:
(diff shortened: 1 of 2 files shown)


kierdavis:
After adding a commit to change the CI config as per #257, CI now fails as expected.

kierdavis:
#257 is merged so this is no longer needed.

Issue #263, opened 2019-04-25
Identifier should be accessible on components.


trickeydan:

trickeydan:
Will need fixing in `test_motor_board.py`

PR #264, opened 2019-04-25
Tweak motor board test script.

trickeydan:


kierdavis:
Accessing a private attribute (`_identifier`) is kinda ugly, but there's nothing we can do about it until #263 is resolved.

PR #279, opened 2019-05-05
Ensure that the run LED is on after the start button flash.

### LINKED ISSUE #274: SR Power Board run led in unknown state after start button flash is pressed.
We need to explicitly turn it on / off.

trickeydan:
Fixes #274

### DIFF
--- a/j5/boards/sr/v4/power_board.py
+++ b/j5/boards/sr/v4/power_board.py
@@ -108,7 +108,11 @@ class PowerBoard(Board):
         self._output_group.power_off()
 
     def wait_for_start_flash(self) -> None:
-        """Wait for the start button to be pressed and flash."""
+        """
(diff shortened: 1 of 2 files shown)


j5/boards/sr/v4/power_board.py
```diff
@@ -108,7 +108,11 @@ def make_safe(self) -> None:
         self._output_group.power_off()
 
     def wait_for_start_flash(self) -> None:
-        """Wait for the start button to be pressed and flash."""
+        """
+        Wait for the start button to be pressed and flash.
+
+        The LED will remain on at the end of the flashing.
```
sedders123:
```suggestion
        The LED will remain on once the start button has been pressed.
```
Would this provide more context? The original could be read as "the LED will flash until it doesn't"

PR #281, opened 2019-05-05
Update links to SR Hardware

trickeydan:
These have been moved to GitHub!


kierdavis:
I think this should also be upstreamed to the SR docs.

Issue #66, opened 2019-02-08
Support SourceBots Arduino


trickeydan:
Requires:
- Servos - #42 
- GPIO - #43 
- Ultrasound - This may or may not want to be a component. Consider this when we get here.

kierdavis:
I think canonically ultrasound sensors *should* be a component; however this may require a rethink of the component API due to how ultrasound sensors are addressed as a pair of pin numbers (trigger pin and echo pin).

trickeydan:
For ultrasound sensors, we could have a 'virtual' `Component`. Two `GPIOPin`s can be passed to it in the constructor?

kierdavis:
yeah. There's no reason why we couldn't have high-level components built on top of low-level components. I'm thinking something like:

```python
class ServoAssembly:
  ...
  def ultrasound_sensor(self, trigger_pin, echo_pin):
    if isinstance(trigger_pin, int): trigger_pin = self.gpio[trigger_pin]
    if isinstance(echo_pin, int): echo_pin = self.gpio[echo_pin]
    return UltrasoundProximitySensor(self._backend, trigger_pin, echo_pin)
  ...
class UltrasoundProximitySensor:
  @property
  def distance(self):
    return self._backend.read_ultrasound_sensor(self._trigger_pin.identifier, self._echo_pin.identifier)
```

kierdavis:
Ultrasound isn't implemented in the hardware backend yet. Fortunately this is just a matter of extending the existing `SBArduinoHardwareBackend` with `UltrasoundInterface`; pretty much all of the infrastructure is already in place.

kierdavis:
#365 fixed this.

Issue #65, opened 2019-02-08
Support SR Ruggeduino


trickeydan:
Requires GPIO Pin -  #43 

We need to think about how to support direct communication. Perhaps some sort of serial component?

kierdavis:
By direct communication, I assume you mean exposing a channel for competitors' robot code to talk to their custom ruggeduino firmware?

trickeydan:
> By direct communication, I assume you mean exposing a channel for competitors' robot code to talk to their custom ruggeduino firmware?

Exactly that.

trickeydan:
I think this probably needs breaking up a bit.

- [ ] Generalisation of Arduino-style boards
- [ ] Generalisation of ArduinoBackend
- [ ] Work out how to send custom commands to the Ruggeduino (I think a new Component is in order here)
- [ ] Implementation of RuggeduinoBoard and Backends.

There are three states that a Ruggeduino can be in:

- Standard SR firmware.
- Extended SR firmware (i.e enable command component)
- Completely custom firmware
	- `sr-robot` has an ignore option for this. We should probably add one of these too.
	- Expose a list of ignored serial devices so that students can use them.

PR #293, opened 2019-05-18
Add minimal board definition for Arduino Uno

trickeydan:
This is useful as the parent class for #66 and #65 

Missing from this board definition:

- PWM on Digital Pins
- I2C
- SPI

These are not in scope for this. This PR only implements what is necessary for #65 and #66. Issues should be added to implement this support in the future.

### DIFF
--- a/j5/boards/arduino/uno.py
+++ b/j5/boards/arduino/uno.py
@@ -1,0 +1,114 @@
+"""Classes for the Arduino Uno."""
+from enum import IntEnum
+from typing import Mapping, Optional, Set, Type, Union, cast
+
+from j5.backends import Backend
+from j5.boards import Board
+from j5.components import (
+    LED,
+    Component,
+    GPIOPin,
+    GPIOPinInterface,
+    GPIOPinMode,
+    LEDInterface,
+)
+
+
+class AnaloguePin(IntEnum):
+    """Analogue Pins numbering."""
+
+    A0 = 14
+    A1 = 15
+    A2 = 16
+    A3 = 17
+    A4 = 18
+    A5 = 19
+
+
+PinNumber = Union[int, AnaloguePin]
+
+
+class ArduinoUnoBoard(Board):
+    """Arduino Uno Board."""
+
+    _led: LED
+    _digital_pins: Mapping[int, GPIOPin]
+    _analogue_pins: Mapping[AnaloguePin, GPIOPin]
+
+    def __init__(self, serial: str, backend: Backend):
+        self._serial = serial
+        self._backend = backend
+
+        self.setup_arduino()
+
+    def setup_arduino(self) -> None:
+        """Setup the arduino components."""
+        self._led = LED(0, cast(self._backend, LEDInterface))  # type: ignore
+
+        # Digital Pins
+        # Note that pins 0 and 1 are used for serial comms.
+        self._digital_pins = {
+            i: GPIOPin(
+                i,
+                cast(GPIOPinInterface, self._backend),
+                initial_mode=GPIOPinMode.DIGITAL_INPUT,
+                supported_modes={
+                    GPIOPinMode.DIGITAL_INPUT,
+                    GPIOPinMode.DIGITAL_INPUT_PULLUP,
+                    GPIOPinMode.DIGITAL_OUTPUT,
+                },
+            )
+            for i in range(2, 14)
+        }
+
+        self._analogue_pins = {
+            i: GPIOPin(
+                i,
+                cast(GPIOPinInterface, self._backend),
+                initial_mode=GPIOPinMode.ANALOGUE_INPUT,
+                supported_modes={
+                    GPIOPinMode.ANALOGUE_INPUT,
+                    GPIOPinMode.DIGITAL_INPUT,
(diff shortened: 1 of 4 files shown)


j5/boards/arduino/uno.py
```diff
@@ -0,0 +1,114 @@
...
+
+    A0 = 14
+    A1 = 15
+    A2 = 16
+    A3 = 17
+    A4 = 18
+    A5 = 19
+
+
+PinNumber = Union[int, AnaloguePin]
```
kierdavis: Reckon we should make a `DigitalPin` enum too, since there's a finite number of them and pins `0` and `1` are invalid? Or do you think that's more trouble than it's worth? I don't have a strong opinion on this.

trickeydan: > pins 0 and 1 are invalid

I'd argue that's a implementation details that should throw an exception in the backend. e.g a different backend / comms protocol could be using i2c and thus block `A4` and `A5`.

> should make a DigitalPin enum too

In principle, I like this suggestion. My one concern is what we would call the entries?

e.g

We can't use `4` as that is a protected thingy in Python.

Maybe `D2` - `D13`? Although that then doesn't match the numbering on the board.

kierdavis: > I'd argue that's a implementation details that should throw an exception in the backend. e.g a different backend / comms protocol could be using i2c and thus block `A4` and `A5`.

Didn't think of this, yeah I agree.

> Maybe `D2` - `D13`?

This is what I was thinking too, but yeah the labelling mismatch might cause confusion.

Maybe we don't need to go the full mile here - we could just use a [type alias](https://mypy.readthedocs.io/en/latest/kinds_of_types.html#type-aliases) for `int` rather than defining a new enum.

trickeydan: > Maybe we don't need to go the full mile here - we could just use a [type alias](https://mypy.readthedocs.io/en/latest/kinds_of_types.html#type-aliases) for `int` rather than defining a new enum.

What's your opinion on using a [NewType](https://mypy.readthedocs.io/en/latest/more_types.html)?

kierdavis: > What's your opinion on using a [NewType](https://mypy.readthedocs.io/en/latest/more_types.html)?

That would be even better. (It was also my original intention, but I wasn't sure whether they existed in mypy and failed to find it in the docs)

trickeydan: I've attempted this and it would require students to do `pins[DigitalPin(4)]`, which is kinda nasty UX, so let's not.

j5/boards/arduino/uno.py
```diff
@@ -0,0 +1,114 @@
...
+
+    def __init__(self, serial: str, backend: Backend):
+        self._serial = serial
+        self._backend = backend
+
+        self.setup_arduino()
+
+    def setup_arduino(self) -> None:
+        """Setup the arduino components."""
+        self._led = LED(0, cast(self._backend, LEDInterface))  # type: ignore
```
kierdavis: Why is `# type: ignore` needed here?

trickeydan: 44f0623

docs/conf.py
```diff
@@ -15,9 +15,10 @@
 import os
 import sys
 
+from j5 import __version__, __version_short__
```
sedders123: Does this break docs generation? Pretty sure it was after the `sys.path.insert()` so that python can be told what `j5` is

j5/boards/arduino/uno.py
```diff
@@ -0,0 +1,114 @@
...
+
+    _led: LED
+    _digital_pins: Mapping[int, GPIOPin]
+    _analogue_pins: Mapping[AnaloguePin, GPIOPin]
+
+    def __init__(self, serial: str, backend: Backend):
+        self._serial = serial
+        self._backend = backend
+
+        self.setup_arduino()
```
sedders123: If we're calling this upon instantiation should it be marked as private, or is it intended to be called multiple times?

trickeydan: This is intended to be called by classes which inherit from this one, (e.g `ServoAssembly`, `Ruggeduino`), but it probably can be private for this purpose anyway.

docs/conf.py
trickeydan: I'm not even sure why this is in this PR...

I'll fix it

j5/boards/arduino/uno.py
kierdavis: Would `super().__init__(serial, backend)` not be a more Pythonic solution?

kierdavis: Specifically:

```python
class ArduinoUnoBoard(Board):
  def __init__(self, serial, backend):
    self._serial = serial
    self._backend = backend
    # generic arduino initialisation goes here
class ServoAssembly(ArduinoUnoBoard):
  def __init__(self, serial, backend):
    super().__init__(serial, backend)
    # servo assembly-specific initialisation goes here
```

trickeydan: 8a0bf73

trickeydan:
I've done a rebase to clean up the history here, but it should be good to go now.

PR #299, opened 2019-06-06
Ultrasound Support

trickeydan:
Resolves: #231, #66

PR #306, opened 2019-06-09
Console Backend for Arduino Uno.

trickeydan:
~Change the base to master once #299 is merged.~

### DIFF
--- a/j5/backends/console/arduino/uno.py
+++ b/j5/backends/console/arduino/uno.py
@@ -81,7 +81,7 @@ class ArduinoUnoConsoleBackend(GPIOPinInterface, LEDInterface, Backend):
         ]:
             raise ValueError(f"Pin {identifier} mode needs to be DIGITAL_INPUT_*"
                              f"in order to read the digital state.")
-        return self._console.read(f"Pin {identifier} digital state [bool]", bool)
+        return self._console.read(f"Pin {identifier} digital state [true/false]", bool)
 
     def read_gpio_pin_analogue_value(self, identifier: int) -> float:
         """Read the scaled analogue value of the GPIO pin."""
@@ -103,11 +103,11 @@ class ArduinoUnoConsoleBackend(GPIOPinInterface, LEDInterface, Backend):
     def get_led_state(self, identifier: int) -> bool:
         """Get the state of an LED."""
         if identifier != 0:
-            raise ValueError("Arduino Uno only has LED 0 (D13).")
+            raise ValueError("Arduino Uno only has LED 0 (digital pin 13).")
         return self.get_gpio_pin_digital_state(13)
 
     def set_led_state(self, identifier: int, state: bool) -> None:
         """Set the state of an LED."""
         if identifier != 0:
-            raise ValueError("Arduino Uno only has LED 0 (D13)")
+            raise ValueError("Arduino Uno only has LED 0 (digital pin 13)")
         self.write_gpio_pin_digital_state(13, state)
--- a/j5/backends/console/arduino/uno.py
+++ b/j5/backends/console/arduino/uno.py
@@ -50,7 +50,7 @@ class ArduinoUnoConsoleBackend(GPIOPinInterface, LEDInterface, Backend):
 
     def set_gpio_pin_mode(self, identifier: int, pin_mode: GPIOPinMode) -> None:
         """Set the hardware mode of a GPIO pin."""
-        self._console.info(f"Set pin {identifier} to {pin_mode}")
+        self._console.info(f"Set pin {identifier} to {pin_mode.name}")
         self._pins[identifier].mode = pin_mode
 
     def get_gpio_pin_mode(self, identifier: int) -> GPIOPinMode:
@@ -93,21 +93,21 @@ class ArduinoUnoConsoleBackend(GPIOPinInterface, LEDInterface, Backend):
     def write_gpio_pin_dac_value(self, identifier: int, scaled_value: float) -> None:
         """Write a scaled analogue value to the DAC on the GPIO pin."""
         # Uno doesn't have any of these.
-        raise NotImplementedError  # pragma: nocover
+        raise NotImplementedError
 
     def write_gpio_pin_pwm_value(self, identifier: int, duty_cycle: float) -> None:
         """Write a scaled analogue value to the PWM on the GPIO pin."""
         # Not implemented on ArduinoUnoBoard yet.
-        raise NotImplementedError  # pragma: nocover
+        raise NotImplementedError
 
     def get_led_state(self, identifier: int) -> bool:
         """Get the state of an LED."""
-        if identifier != 13:
-            raise ValueError("Arduino Uno only has an LED on pin D13.")
-        return self.read_gpio_pin_digital_state(13)
+        if identifier != 0:
+            raise ValueError("Arduino Uno only has LED 0 (D13).")
+        return self.get_gpio_pin_digital_state(13)
 
     def set_led_state(self, identifier: int, state: bool) -> None:
         """Set the state of an LED."""
-        if identifier != 13:
-            raise ValueError("Arduino Uno only has an LED on pin D13.")
+        if identifier != 0:
+            raise ValueError("Arduino Uno only has LED 0 (D13)")
         self.write_gpio_pin_digital_state(13, state)
--- a/j5/backends/console/arduino/uno.py
+++ b/j5/backends/console/arduino/uno.py
@@ -1,0 +1,113 @@
+"""Console Backend for the Arduino Uno."""
+from typing import Mapping, Optional, Set, Type
+
+from j5.backends import Backend
+from j5.backends.console import Console, ConsoleEnvironment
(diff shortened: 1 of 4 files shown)


j5/backends/console/arduino/uno.py
```diff
@@ -0,0 +1,113 @@
...
+
+    def get_led_state(self, identifier: int) -> bool:
+        """Get the state of an LED."""
+        if identifier != 0:
+            raise ValueError("Arduino Uno only has LED 0 (D13).")
+        return self.get_gpio_pin_digital_state(13)
+
+    def set_led_state(self, identifier: int, state: bool) -> None:
+        """Set the state of an LED."""
+        if identifier != 0:
```
sedders123: `False != 0` evaluates to `False`, because of this If a competitor got the order of the arguments the wrong way around they'd get some unusual behaviour. Is it worth checking that `identifier` is an int and returning an nicer error message if its not?

j5/backends/console/arduino/uno.py
```diff
@@ -0,0 +1,113 @@
...
+        raise NotImplementedError("The Console Backend cannot discover boards.")
+
+    def __init__(self, serial: str, console_class: Type[Console] = Console) -> None:
+        self._serial = serial
+
+        self._pins: Mapping[int, PinData] = {
+            i: PinData(mode=GPIOPinMode.DIGITAL_OUTPUT, digital_state=False)
+            for i in range(2, 20)
+            # Digital 2 - 13
+            # Analogue 14 - 19
```
sedders123: By the looks of things all pins here are set to `GPIOPinMode.DIGITAL_OUTPUT` by default. Should this comment about analogue pin ranges be removed, or am I missing something?

trickeydan: Analogue Pins (A0-A5) on the Arduino Uno also support Digital write and read, so this is correct.

This backend doesn't store the state for the Analogue state of the pins, it is given through stdin.

j5/backends/console/arduino/uno.py
trickeydan: That's taken care of in `j5/components/led.py`. The functions on the backend aren't designed to be called by competitors, and should only contain implementation specific validation.

j5/backends/console/arduino/uno.py
```diff
@@ -0,0 +1,113 @@
...
+    def read_gpio_pin_digital_state(self, identifier: int) -> bool:
+        """Read the digital state of the GPIO pin."""
+        if self._pins[identifier].mode not in [
+            GPIOPinMode.DIGITAL_INPUT_PULLUP,
+            GPIOPinMode.DIGITAL_INPUT,
+            GPIOPinMode.DIGITAL_INPUT_PULLDOWN,
+        ]:
+            raise ValueError(f"Pin {identifier} mode needs to be DIGITAL_INPUT_*"
+                             f"in order to read the digital state.")
+        return self._console.read(f"Pin {identifier} digital state [bool]", bool)
```
kierdavis: Consistency with https://github.com/j5api/j5/blob/290fdbc/j5/backends/console/sr/v4/power_board.py#L110.

```suggestion
        return self._console.read(f"Pin {identifier} digital state [true/false]", bool)
```

j5/backends/console/arduino/uno.py
```diff
@@ -0,0 +1,113 @@
...
+
+    def read_gpio_pin_analogue_value(self, identifier: int) -> float:
+        """Read the scaled analogue value of the GPIO pin."""
+        if self._pins[identifier].mode is not GPIOPinMode.ANALOGUE_INPUT:
+            raise ValueError(f"Pin {identifier} mode needs to be ANALOGUE_INPUT",
+                             f"in order to read the digital state.")
+        return self._console.read(f"Pin {identifier} ADC state [float]", float)
```
kierdavis: This prompt should say something more useful than "float", but exactly what depends on #315.

j5/backends/console/arduino/uno.py
```diff
@@ -0,0 +1,113 @@
...
+
+    def write_gpio_pin_pwm_value(self, identifier: int, duty_cycle: float) -> None:
+        """Write a scaled analogue value to the PWM on the GPIO pin."""
+        # Not implemented on ArduinoUnoBoard yet.
+        raise NotImplementedError
+
+    def get_led_state(self, identifier: int) -> bool:
+        """Get the state of an LED."""
+        if identifier != 0:
+            raise ValueError("Arduino Uno only has LED 0 (D13).")
```
kierdavis: I think we agreed not to use the `D<number>` convention since it doesn't match the labels on the board.

```suggestion
            raise ValueError("Arduino Uno only has LED 0 (digital pin 13).")
```

j5/backends/console/arduino/uno.py
```diff
@@ -0,0 +1,113 @@
...
+        return self.get_gpio_pin_digital_state(13)
+
+    def set_led_state(self, identifier: int, state: bool) -> None:
+        """Set the state of an LED."""
+        if identifier != 0:
+            raise ValueError("Arduino Uno only has LED 0 (D13)")
```
kierdavis: I think we agreed not to use the `D<number>` convention since it doesn't match the labels on the board.

```suggestion
            raise ValueError("Arduino Uno only has LED 0 (digital pin 13)")
```

tests/backends/console/arduino/test_uno.py
```diff
@@ -0,0 +1,228 @@
...
+
+
+def test_set_gpio_pin_mode() -> None:
+    """Test that we can set the mode of a GPIO pin."""
+    backend = ArduinoUnoConsoleBackend(
+        "TestBoard",
+        console_class=MockConsole,
+    )
+
+    backend._console.expects = "Set pin 10 to DIGITAL_OUTPUT"  # type: ignore
```
kierdavis: I'm guessing `type: ignore` is needed here because the typechecker doesn't know that `backend._console` is a `MockConsole` rather than the more general `Console`. Could we instead pull out the console into a local variable using a cast and refer to it directly?

```python
console = cast(MockConsole, backend._console)
console.expects = ...
```

Even better, could we refactor so that the console object is created in this function and passed into the `ArduinoUnoConsoleBackend` constructor, rather than being created inside the constructor and having to be pulled out of a private attribute?

tests/backends/console/arduino/test_uno.py
```diff
@@ -0,0 +1,228 @@
...
+        console_class=MockConsole,
+    )
+
+    backend._console.expects = "Set pin 13 state to False"  # type: ignore
+    backend.set_led_state(0, False)
+    assert not backend.get_led_state(0)
+
+    backend._console.expects = "Set pin 13 state to True"  # type: ignore
+    backend.set_led_state(0, True)
+    assert backend.get_led_state(0)
```
kierdavis: Could we make this check the LED and digital pin 13 are the same thing? i.e. write using `set_led_state(0, ...)` and read back using `get_gpio_pin_digital_state(13)`, and vice versa.

tests/backends/console/arduino/test_uno.py
trickeydan: I've copied how this is done in other `ConsoleBackend` implementations. I agree that your method is much better.

We should refactor this everywhere at once in my opinion, but that is out of scope for this PR.

j5/backends/console/arduino/uno.py
kierdavis:
```suggestion
        return self._console.read(f"Pin {identifier} analogue voltage [volts]", float)
```

tests/backends/console/arduino/test_uno.py
kierdavis: Sounds good to me.

j5/backends/console/arduino/uno.py
```diff
@@ -0,0 +1,113 @@
...
+from typing import Mapping, Optional, Set, Type
+
+from j5.backends import Backend
+from j5.backends.console import Console, ConsoleEnvironment
+from j5.boards import Board
+from j5.boards.arduino import ArduinoUnoBoard
+from j5.components import GPIOPinInterface, GPIOPinMode, LEDInterface
+
+
+class PinData:
```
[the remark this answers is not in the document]
trickeydan: Unfortunately this is not possible afaik.

`NamedTuple` has immutable properties and we need to modify them in this Backend.

An alternative is [`recordclass`](https://pypi.python.org/pypi/recordclass), but this would add another dependency, so I'd rather use a class.

tests/backends/console/arduino/test_uno.py
trickeydan: b7e8229

j5/backends/console/arduino/uno.py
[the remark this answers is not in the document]
kierdavis: I think [`namedtuple._replace`](https://docs.python.org/3/library/collections.html#collections.somenamedtuple._replace) is the canonical way to "modify" a named tuple, but that's probably an uglier solution. Lets leave this as it is.

PR #309, opened 2019-06-10
Uncomment servo init

### LINKED ISSUE #308: Commented-out line in Servo Backend needs investigating.
https://github.com/j5api/j5/blob/7c90ee9886ddf9fbc88ca942639e94aa53611a50/j5/backends/hardware/sr/v4/servo_board.py#L65

I believe this was due to them crashing every time I tried with this line in there, despite it being in the protocol. Possibly a side-effect of the buggy firmware.

trickeydan:
Fixes #308

### DIFF
--- a/j5/backends/hardware/sr/v4/servo_board.py
+++ b/j5/backends/hardware/sr/v4/servo_board.py
@@ -62,7 +62,7 @@ class SRV4ServoBoardHardwareBackend(
 
         # Initialise servos.
 
-        # self._usb_device.ctrl_transfer(0, 64, 0, 12, b"")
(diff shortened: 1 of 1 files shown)


kierdavis:
I'd like to confirm this myself before approving, since I thought we concluded pretty much the opposite (that it only worked *without* this line) back when this was written.

trickeydan:
The behaviour is inconsistent at the very least

kierdavis:
Only works for me (on a laptop) with the line uncommented.

trickeydan:
Given that it needs this line on both the Pi I tested last week and my laptop now.

I suspect the original inconsistent behaviour may have been caused by a dev board that had modified firmware..

PR #323, opened 2019-06-30
Marker / Coordinate Structures

HU90m:
An initial marker class concept.


j5/vision/markers.py
```diff
@@ -0,0 +1,203 @@
...
+    def add_markers(self, given_positions: List[Coordinate]) -> None:
+        """Add marker(s) by providing their coordinates."""
+        given_positions = np.array([position.to_cartesian()
+                                    for position in given_positions
+                                    ])
+        self.__append_positions(given_positions)
+
+    def __append_positions(self, given_positions):
+        """Appends a numpy array of markers to the current array (if one exists)."""
+        if self.__positions is not None:
```
RealOrangeOne: if `self.__positions` is initialized as an empty list, this dance isn't necessary.

j5/vision/markers.py
```diff
@@ -0,0 +1,203 @@
...
+            self.__positions = np.concatenate([self.__positions, given_positions])
+        else:
+            self.__positions = given_positions
+
+    def translate_space(self, delta_x, delta_y, delta_z):
```
RealOrangeOne: Why are we supporting mutation of the markers? Markers represent a position in physical space, why can't be mutated by the robot in the way these methods imply.

j5/vision/markers.py
HU90m: That is a far better solution. I will implement it.

j5/vision/markers.py
```diff
@@ -0,0 +1,203 @@
...
+class Cartesian(NamedTuple):
+    """A Cartesian Coordinate."""
+
+    x: float
+    y: float
+    z: float
+
+    def as_cartesian(self) -> Cartesian:
+        """Provides the coordiante in the form of a cartesian coordinate."""
+        return self
```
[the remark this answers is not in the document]
HU90m: I added this for for add_markers method so that it was coordinate system agnostic. I would also like to be able to return coordinates using any system, so that teams can use the coordinate system that best suits them and the control logic.

I plan to the remove the markers class' static conversion methods, because these are made redundant by the casting interface and I don't think the convertions should be handled by the markers class.

j5/vision/markers.py
```diff
@@ -0,0 +1,203 @@
...
+
+    def as_cylindrical(self) -> Cylindrical:
+        """Provides the coordiante in the form of a cylindrical coordinate."""
+        return Cylindrical(np.sqrt(self.x**2 + self.y**2),
+                           np.arctan2(self.y, self.x),
+                           self.z)
+
+    def as_spherical(self) -> Spherical:
+        """Provides the coordiante in the form of a spherical coordinate."""
+        return Spherical(np.sqrt(self.x**2 + self.y**2 + self.z**2),
```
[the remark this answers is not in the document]
HU90m: linalg.norm() does look perfect for this. I will have a look for a python built-in equivalent method.

j5/vision/markers.py
```diff
@@ -0,0 +1,203 @@
...
+    z   := height
+    """
+
+    p: float
+    phi: float
+    z: float
+
+    def as_cartesian(self) -> Cartesian:
+        """Provides the coordiante in the form of a cartesian coordinate."""
+        return Cartesian(self.p * np.cos(self.phi),
```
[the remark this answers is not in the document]
HU90m: These are these are relics from when the named tuples used to contain numpy arrays. Moving back to built-in methods for the named tuples is a great idea.

j5/vision/markers.py
```diff
@@ -0,0 +1,203 @@
...
+    The second dimension is the axis.
+    Example:
+        [ [x1, y1, z1],
+          [x2, y2, z2],
+          [x3, y3, z3],
+          [x4, y4, z4] ]
+    """
+
+    def __init__(self):
+        self.__positions = None
```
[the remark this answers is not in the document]
HU90m: What metadata is required for the j5 markers?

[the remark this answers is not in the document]
trickeydan: Pretty much just a number.

Take a look at previous APIs I linked in the issue :)

j5/vision/markers.py
```diff
@@ -0,0 +1,203 @@
...
+    def positions(self) -> List[Cartesian]:
+        """Returns the makers positions."""
+        return [Cartesian(*coordinate) for coordinate in self.__positions]
+
+    def add_markers_cartesian(self, given_positions: List[Cartesian]) -> None:
+        """Add marker(s) by providing their cartesian coordinates."""
+        given_positions = np.array(given_positions)
+        self.__append_positions(given_positions)
+
+    def add_markers(self, given_positions: List[Coordinate]) -> None:
```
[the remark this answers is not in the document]
HU90m: Yeah, I was going to call the add_markers() method in the constructor.

j5/vision/markers.py
HU90m: My plan was to keep the robot at the origin. The mutations allow the robot to guess where the markers will be after a movement without having to wait for the vision system to update marker locations. This means robots speed isn't limited by the vision system.

j5/vision/markers.py
```diff
@@ -0,0 +1,203 @@
...
+        self.__positions = None
+
+    @property
+    def positions(self) -> List[Cartesian]:
+        """Returns the makers positions."""
+        return [Cartesian(*coordinate) for coordinate in self.__positions]
+
+    def add_markers_cartesian(self, given_positions: List[Cartesian]) -> None:
+        """Add marker(s) by providing their cartesian coordinates."""
+        given_positions = np.array(given_positions)
```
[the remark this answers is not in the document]
HU90m: I chose to use numpy arrays because they allow the fast mutatation of the coordinate space and additional markers become less expensive (speed, not space).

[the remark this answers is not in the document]
sedders123: Has this been measured or is the speed increase an assumption? I'm keen to keep to the standard library as much as possible unless significant performance benefits can be demonstrated.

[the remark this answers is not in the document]
HU90m: Yeah, its an assumption. I will get rid of the numpy dependancy. We can look into using numpy in the future if we need to improve the performance.

j5/vision/markers.py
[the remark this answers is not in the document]
HU90m: Oh brilliant, thats easy.

j5/vision/markers.py
HU90m: Whats the verdict? Are space transforms wanted?

trickeydan: I personally don't like the markers being mutable. I think we can implement a function that maps over a MarkerList to return a new list quite easily also.

j5/vision/markers.py
```diff
@@ -0,0 +1,129 @@
...
+
+    __position: Coordinate
+    __id: int
+
+    def __init__(self, id: int, position: Coordinate):
+        self.__id = id
+        self.__position = position
+
+    @property
+    def id(self) -> List[Coordinate]:
```
trickeydan: Pretty sure that an ID is an int?

Can we add a test for this too?

tests/vision/test_markers.py
```diff
@@ -0,0 +1,136 @@
+"""Tests for the Marker Class."""
+
+from math import isclose
+
+from j5.vision.markers import Coordinate, Cylindrical, Marker, Spherical
+
+
+def assert_coordinate_isclose(coord1: Coordinate, coord2: Coordinate):
+    """Checks two coordinates a roughly equal."""
```
trickeydan:
```suggestion
    """Checks two coordinates are roughly equal."""
```

tests/vision/test_markers.py
```diff
@@ -0,0 +1,136 @@
...
+
+def assert_cylindrical_isclose(cyl1: Cylindrical, cyl2: Cylindrical):
+    """Checks two cylindricals a roughly equal."""
+    assert isclose(cyl1.p, cyl2.p)
+    assert isclose(cyl1.theta, cyl2.theta)
+    assert isclose(cyl1.z, cyl2.z)
+
+
+def assert_spherical_isclose(sph1: Spherical, sph2: Spherical):
+    """Checks two sphericals a roughly equal."""
```
trickeydan:
```suggestion
    """Checks two sphericals are roughly equal."""
```

tests/vision/test_markers.py
```diff
@@ -0,0 +1,136 @@
...
+
+def assert_coordinate_isclose(coord1: Coordinate, coord2: Coordinate):
+    """Checks two coordinates a roughly equal."""
+    assert isclose(coord1.cartesian.x, coord2.cartesian.x)
+    assert isclose(coord1.cartesian.y, coord2.cartesian.y)
+    assert isclose(coord1.cartesian.z, coord2.cartesian.z)
+
+
+def assert_cylindrical_isclose(cyl1: Cylindrical, cyl2: Cylindrical):
+    """Checks two cylindricals a roughly equal."""
```
trickeydan:
```suggestion
    """Checks two cylindricals are roughly equal."""
```

j5/vision/markers.py
```diff
@@ -0,0 +1,129 @@
...
+
+    r: float
+    theta: float
+    phi: float
+
+
+class Marker():
+    """Marker Class.
+
+    Holds a maker's position and id.
```
kierdavis:
```suggestion
    Holds a marker's position and id.
```

j5/vision/markers.py
```diff
@@ -0,0 +1,129 @@
...
+    """
+
+    __position: Coordinate
```
kierdavis: [Python treats attributes beginning with two underscores specially](https://docs.python.org/3/tutorial/classes.html#private-variables). Is such behaviour needed here, or would using a single underscore be sufficient?

tests/vision/test_markers.py
```diff
@@ -0,0 +1,136 @@
+"""Tests for the Marker Class."""
+
+from math import isclose
+
+from j5.vision.markers import Coordinate, Cylindrical, Marker, Spherical
+
+
+def assert_coordinate_isclose(coord1: Coordinate, coord2: Coordinate):
```
kierdavis: Just a thought: I wonder if creating `Cartesian.isclose`/`Cylindrical.isclose`/`Spherical.isclose` methods would be useful to users. Then this function could be replaced with simply `assert coord1.cartesian.isclose(coord2.cartesian)`.

tests/vision/test_markers.py
```diff
@@ -0,0 +1,136 @@
...
+
+    cylindrical_1 = Coordinate.from_cylindrical(5., 0.92729521800161, 5.)
+    cylindrical_2 = Coordinate.from_cylindrical(20.223748416157, 0.1488899476095, 7.)
+
+    assert_coordinate_isclose(coordinate_1, cylindrical_1)
+    assert_coordinate_isclose(coordinate_2, cylindrical_2)
+
+
+def test_coordinate_from_spherical() -> None:
+    """Test the from_spherical() method of the coordiante class."""
```
kierdavis:
```suggestion
    """Test the from_spherical() method of the coordinate class."""
```

tests/vision/test_markers.py
```diff
@@ -0,0 +1,136 @@
...
+
+    assert_spherical_isclose(coordinate_1.spherical, spherical_1)
+    assert_spherical_isclose(coordinate_2.spherical, spherical_2)
+
+
+def test_coordinate_represenation() -> None:
+    """Test the representation of the coordinate class."""
+    coordinate_1 = Coordinate(3., 4., 5.)
+
+    given_representation = coordinate_1.__repr__()
```
kierdavis: I think this is a more conventional way to do this:

```suggestion
    given_representation = repr(coordinate_1)
```

j5/vision/markers.py
trickeydan: I personally would rather that this was `_position`

trickeydan:
Let's **not** include the following in this PR to try and keep it to a reasonable size:

- Orientation
- Camera Component
- Integration with Zoloto

i.e Anything that isn't Marker / Coordinates

j5/vision/markers.py
trickeydan: See #341 where I made a fix for this

tests/vision/test_markers.py
trickeydan: See #341 8ae1d3c where I have implemented `Coordinate.isclose`

trickeydan:
Rebased and cleaned up

trickeydan:
The Github GUI on the main page seems confused, look in the commits tab please!

PR #326, opened 2019-07-08
Various fixes and tweaks to vision

trickeydan:


j5/vision/markers.py
```diff
@@ -1,35 +1,59 @@
...
 
+class Coordinate(metaclass=ABCMeta):
+    """A position in space."""
+
+    _cart: 'Cartesian'
+    _cyl: 'Cylindrical'
+    _sph: 'Spherical'
+
+    def __init__(self, x: float, y: float, z: float):
+        self._cart = Cartesian(x, y, z)
```
HU90m: Could the constructor be able to take cylindrical and spherical coordinates as well as catesian coordinates?

PR #341, opened 2019-07-16
Finish vision constructs

trickeydan:
Finishing off the initial vision data constructs. 

Orientation is deliberately excluded at this point, as it is not a MUST for a shippable vision system.


trickeydan:
I've done a merge from master into this branch. Ignore any commits before 
efc21ab

RealOrangeOne:
This PR is almost impossible to review, as there's so many other unrelated commits in the diff. Can we get the unrelated commits removed from the branch, or just wait until the other branches are merged?

trickeydan:
I'm unable to get that diff nice, so I think the best approach will be to merge this into `vision` without review, and then review the diff into `master`. This approach will also prevent duplicated comments on these rather big changes.

trickeydan:
I'm messaging @HU90m directly to inform him that I'm pushing to his branch as that would generally not be considered best practice.

Issue #301, opened 2019-06-09
Zoloto Support


trickeydan:
We should support `zoloto` as a vision library. It's awesome.

https://github.com/realorangeone/zoloto

PR #345, opened 2019-07-21
Fiducial Marker Detection

trickeydan:


trickeydan:
I'm aware the test coverage is almost non-existent. I think that there are high priorities currently. We should still do it asap though, just not in this PR

trickeydan:
Closes #301

j5/backends/hardware/zoloto/camera_board.py
```diff
@@ -0,0 +1,100 @@
...
+from j5.backends.hardware import HardwareEnvironment
+from j5.boards import Board
+from j5.boards.zoloto import ZolotoCameraBoard
+from j5.components import MarkerCameraInterface
+from j5.vision import Coordinate, Marker, MarkerList
+
+CAMERA_PATH = PosixPath("/dev/video0")
+CAMERA_SERIAL = "video0"
+
+T = TypeVar("T", bound=Camera)
```
HU90m: Is 'T' left over from development or does it serve a purpose?

trickeydan: It's used in the backend constructor to allow subclasses of Camera

trickeydan:
Awaiting fix of upstream issue https://github.com/RealOrangeOne/zoloto/issues/41

j5/backends/console/zoloto/camera_board.py
```diff
@@ -0,0 +1,80 @@
...
+                Marker(marker_id, Coordinate(marker_x, marker_y, marker_z)),
+            )
+
+        return markers
+
+    def save_annotated_image(self, file: Path) -> None:
+        """Save an annotated image to a file."""
+        self._console.info(f"Saved annotated image to {file}")
+        self._console.info(
+            f"(Image not actually saved due to backend limitations.",
```
RealOrangeOne: Looks like you've got a stray bracket here, and this probably doesn't need to be an `f`-string

j5/backends/hardware/zoloto/camera_board.py
```diff
@@ -0,0 +1,100 @@
...
+from zoloto.cameras.camera import Camera
+
+from j5.backends import Backend
+from j5.backends.hardware import HardwareEnvironment
+from j5.boards import Board
+from j5.boards.zoloto import ZolotoCameraBoard
+from j5.components import MarkerCameraInterface
+from j5.vision import Coordinate, Marker, MarkerList
+
+CAMERA_PATH = PosixPath("/dev/video0")
```
RealOrangeOne: Is there a reason this is `PosixPath` over simply `Path`?

j5/backends/hardware/zoloto/camera_board.py
```diff
@@ -0,0 +1,100 @@
...
+    board = ZolotoCameraBoard
+
+    @classmethod
+    def discover(cls, camera_class: Type[Camera] = DefaultCamera) -> Set[Board]:
+        """Discover boards that this backend can control."""
+        if system() != "Linux":
+            # We currently only support Zoloto on Linux platforms as there is
+            # no easy way to detect the presence of webcams on other platforms.
+            return set()
+        if not exists(CAMERA_PATH):
```
RealOrangeOne: You can simply call `CAMERA_PATH.exists()` here, which looks much nicer!

j5/backends/hardware/zoloto/camera_board.py
```diff
@@ -0,0 +1,100 @@
...
+    @property
+    def firmware_version(self) -> Optional[str]:
+        """The firmware version reported by the board."""
+        return zoloto_version
+
+    def get_visible_markers(self, identifier: int) -> MarkerList:
+        """Get markers that are visible to the camera."""
+        markers = MarkerList()
+
+        marker_gen = self._zcamera.process_frame()
```
RealOrangeOne: I _highly_ recommend using `process_frame_eager` here, as it only does a pose estimation for each _size_ of marker, rather than for each marker it sees, which is a fair amount faster!

j5/boards/zoloto/camera_board.py
```diff
@@ -0,0 +1,45 @@
...
+
+    @property
+    def serial(self) -> str:
+        """Get the serial number."""
+        return self._serial
+
+    @property
+    def firmware_version(self) -> Optional[str]:
+        """Get the firmware version of the board."""
+        return None
```
RealOrangeOne: You can probably still use the version of `zoloto` here.

j5/boards/zoloto/camera_board.py
```diff
@@ -0,0 +1,45 @@
...
+
+    @property
+    def camera(self) -> MarkerCamera:
+        """The camera on this board."""
+        return self._camera
+
+    def make_safe(self) -> None:
+        """Make this board safe."""
+        pass
```
RealOrangeOne: Here you should call `.close` on the underlying zoloto camera

stubs/zoloto/__init__.py
```diff
@@ -0,0 +1,10 @@
+"""Zoloto."""
+
+from enum import IntEnum
+
+__version__: str
+
+
+class MarkerDict(IntEnum):
```
RealOrangeOne: Why is this necessary? The provided types should be more than enough

trickeydan: Due to the way that mypy stubs work, it will completely ignore any types in your library unfortunately. The stubs can be removed at a later date when upstream has the hints and implements PEP 561

j5/backends/hardware/zoloto/camera_board.py
trickeydan: Mostly because this is a unix path, but am happy to change.

j5/boards/zoloto/camera_board.py
trickeydan: This is a bit of j5 which needs fixing, there is an issue for it #255. Have implemented the correct interim behaviour.

j5/backends/hardware/zoloto/camera_board.py
trickeydan: I'm not bumping this until after the upstream is fixed

Issue #346, opened 2019-07-22
Add tests for Zoloto


trickeydan:
#345 didn't include tests for various reasons.

kierdavis:
Assigning myself (for now), because I've never gotten that involved in vision stuff before and I think this would a good way to dive in.

trickeydan:
This issue should be transferred along with #402

trickeydan:
This is blocked on Zoloto Type Stubs #524

PR #353, opened 2019-07-26
Vision orientations.

trickeydan:


stubs/pyquaternion.pyi
```diff
@@ -0,0 +1,16 @@
+"""Type stubs for PyQuaternion."""
+
+from typing import List, Tuple
+
+class Quaternion:
+
+    # Note __init__ has many possible constructors. This is the one we use.
+    def __init__(self, axis: Tuple[float, float, float], angle: float): ...
```
kierdavis:
```suggestion
    def __init__(self, *, axis: Tuple[float, float, float], angle: float): ...
```

since I think all arguments to `Quarternion` constructors have to be specified as keyword arguments to avoid ambiguity (or at least I think we should enforce this in our code).

j5/vision/orientation.py
```diff
@@ -0,0 +1,84 @@
...
+
+class Orientation:
+    """
+    Represents an orientation in 3D space.
+
+    Uses a unit quaternion as an internal representation.
+    """
+
+    def __init__(self, orientation: Quaternion):
+        self._quaternion = orientation
```
kierdavis: Can we expose the quaternion as a property? I can imagine some competitors may well find a use for it.

```python
@property
def quaternion(self):
    return self._quaternion
```

kierdavis:
No tests, but we're under time constraints. I imagine this code is covered by #346.

j5/vision/orientation.py
```diff
@@ -0,0 +1,84 @@
...
+        return self.yaw_pitch_roll[1]
+
+    @property
+    def roll(self) -> float:
+        """Rotation angle around the x''-axis in radians."""
+        return self.yaw_pitch_roll[2]
+
+    @property
+    def rot_x(self) -> float:
+        """Alias to roll."""
```
sedders123:
```suggestion
        """Returns the rotation around the x axis in radians."""
```

j5/vision/orientation.py
```diff
@@ -0,0 +1,84 @@
...
+        return self.roll
+
+    @property
+    def rot_y(self) -> float:
+        """Alias to pitch."""
```
sedders123:
```suggestion
        """Returns the rotation around the y axis in radians."""
```

j5/vision/orientation.py
```diff
@@ -0,0 +1,84 @@
...
+        return self.pitch
+
+    @property
+    def rot_z(self) -> float:
+        """Alias to yaw."""
```
sedders123:
```suggestion
        """Returns the rotation around the z axis in radians."""
```

PR #355, opened 2019-07-27
Rename motor.state to motor.power

trickeydan:


kierdavis:
And so the search for an accurate name for this property continues...

(LGTM)

trickeydan:
> And so the search for an accurate name for this property continues...
> 
> (LGTM)

I wanted to just add an alias, but @RealOrangeOne was insisting on a rename.

kierdavis:
> I wanted to just add an alias, but @RealOrangeOne was insisting on a rename.

I agree with renaming - an alias might lead to additional confusion for users.

PR #356, opened 2019-07-27
Allow Piezo.buzz to use int and float

trickeydan:


j5/components/piezo.py
```diff
@@ -68,8 +68,14 @@ def identifier(self) -> int:
         """An integer to identify the component on a board."""
         return self._identifier
 
-    def buzz(self, duration: timedelta, pitch: Pitch) -> None:
-        """Queue a note to be played."""
+    def buzz(self, duration: Union[int, float, timedelta], pitch: Pitch) -> None:
+        """
+        Queue a note to be played.
+
+        Float durations are measured in seconds.
```
sedders123:
```suggestion
        Float and integer durations are measured in seconds.
```

kierdavis:
lgtm

PR #359, opened 2019-07-28
Convert Zoloto output to metres

trickeydan:


trickeydan:
Slightly relevant: #346

kierdavis:
Any reason why Zoloto outputs centimetres and not metres in the first place?

trickeydan:
> Any reason why Zoloto outputs centimetres and not metres in the first place?

Zoloto outputs in relative units. Unfortunately the function which sets those relative units can only return an `int` currently.

PR #369, opened 2019-08-06
Add DEBUG-level logging of all Interface method calls

kierdavis:
This uses some funky introspection whenever a Backend is registered to:
* get all of the interfaces it implements
* get all of the abstract methods of these interfaces (component-backend interactions)
* install a wrapper around each method
* use the call signature of the method to get a mapping of the arguments to the call with both names and values
* log the method name, arguments and return value...
* ...to the logger named by the module that the interface is defined in!

This code contains copious amounts of black magic. Perhaps manually
adding log messages to every interface method would have been a more
obvious if less efficient solution.

------

Since writing this I think it might be *too* verbose for typical debugging efforts. Thoughts on defining a new log level below `DEBUG` e.g. `TRACE`?

### DIFF
--- a/j5/backends/backend.py
+++ b/j5/backends/backend.py
@@ -1,6 +1,9 @@
 """The base classes for backends."""
 
+import inspect
+import logging
 from abc import ABCMeta, abstractmethod
+from functools import wraps
 from typing import TYPE_CHECKING, Dict, Optional, Set, Type
 
 if TYPE_CHECKING:  # pragma: nocover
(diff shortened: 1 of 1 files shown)


trickeydan:
This is both nice and not nice.

I'm happy to ship it for now, but I'd like to review whether we will want it in a month

As for verboseness, j5 won't have any other level, so I'm unsure what to do. @sedders123 @RealOrangeOne opinions?

kierdavis:
I've opened #370 to track cleaning this up at a later date.

PR #375, opened 2019-08-12
Set SO_REUSEADDR on the lock socket

kierdavis:
This should prevent the OS from reserving the address for a short
period after the process dies, preventing new processes from
binding to the address.

However I've been unable to reproduce that issue myself, so I've got no idea if it solves the issue. I would not recommend merging until this is confirmed.


kierdavis:
This is not actually an issue, because we don't listen on the socket. We thought it was an issue, but it actually something else (a user putting `r = Robot()` inside a while loop). `SO_REUSEADDR` actually makes the situation worse by letting multiple processes bind to the same socket.

PR #399, opened 2019-08-30
Test Coverage is Gud

trickeydan:


tests/backends/console/test_env.py
```diff
@@ -54,3 +54,31 @@ def mock_input(prompt: str) -> str:
 
     console = Console("TestBoard", input_function=mock_input)
     assert console.read("Enter test input", None) is None
+
+
+BAD_ATTEMPT_COUNT = 0
```
kierdavis: I think it would be better practice to move `mock_input` and `mock_print` into a class, make `BAD_ATTEMPT_COUNT` an attribute of the class, and pass the functions into `Console()` as bound methods.

```python
def test_console_read_bad_type() -> None:
  class Mock:
    def __init__(self):
      self.bad_attempt_count = 0
    def input(self, prompt: str) -> str:
      ...
    def print(self, text: str) -> None:
      ...
  mock = Mock()
  console = Console(
    "TestConsole",
    print_function=mock.print,
    input_function=mock.input,
  )
  ...
```

This pattern is strongly suggesting to me that the `Console` class isn't very pythonic - changing functionality should be done by overriding in a subclass, not by passing functions into its constructor.

j5/components/piezo.py
```diff
@@ -31,12 +31,6 @@ class Note(float, Enum):
     B7 = 3951.1
     C8 = 4186.0
 
-    def __reverse__(self) -> Generator['Note', None, None]:
```
kierdavis: :+1:

I don't think it was useful to have this function in our API anyway. On the offchance that someone needs a list of notes in reverse order, they can construct that sequence themselves quite easily.

trickeydan: It actually works anyway. You can run `reversed(Note)` and it will iterate in reverse order. Hence why the method was removed but not the test.

kierdavis: Ah, looks like this is just a side effect of enums being iterable:

```
Python 3.7.3 (default, Mar 25 2019, 20:59:09) 
[GCC 7.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from enum import Enum, auto
>>> class MyEnum(Enum):
...   A = auto()
...   B = auto()
...   C = auto()
... 
>>> list(MyEnum)
[<MyEnum.A: 1>, <MyEnum.B: 2>, <MyEnum.C: 3>]
>>> list(reversed(MyEnum))
[<MyEnum.C: 3>, <MyEnum.B: 2>, <MyEnum.A: 1>]
```

trickeydan:
I was planning to cover Zoloto in this PR also, but in lieu of #402, I'm not going to bother for now.

PR #400, opened 2019-09-01
Remove typing extensions import check

### LINKED ISSUE #396: Technical Debt around Protocol
```
if TYPE_CHECKING: 

    from typing_extensions import Protocol  # pragma: nocover 

else: 

    class Protocol: 

        """Dummy class since typing_extensions is not available at runtime.""" 

 

        pass 
```

We should see if there is a better way to do this.

`j5/backends/j5/serial.py`

kierdavis:
Fixes #396.


j5/backends/hardware/j5/serial.py
```diff
@@ -2,23 +2,16 @@
 from abc import abstractmethod
 from datetime import timedelta
 from functools import wraps
-from typing import TYPE_CHECKING, Callable, Optional, Set, Type, TypeVar
+from typing import Callable, Optional, Set, Type, TypeVar
 
 from serial import Serial, SerialException, SerialTimeoutException
+from typing_extensions import Protocol
```
trickeydan: Have you checked that this line runs OK on 3.8, where typing_extensions isn't present?

kierdavis: Yep:

```
Python 3.8.0b3 (default, Aug 14 2019, 11:44:08) 
[GCC 8.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from typing_extensions import Protocol
>>> Protocol
<class 'typing.Protocol'>
```

Note that as per https://github.com/j5api/j5/issues/396#issuecomment-526909981, I've set this package to always be installed (instead of only for python<3.8).

Issue #363, opened 2019-08-03
opencv-python-contrib should be an optional dependency


kierdavis:
Ideally it should only be installed when the `zoloto-vision` extra is selected.

kierdavis:
Unfortunately this cannot be included in the `zoloto-vision` extra, because it is a dev dependency. [Apparently this is by design.](https://github.com/sdispater/poetry/pull/606#issuecomment-437943927)

It also can't be made a floating optional dependency, because there's no way to install an optional dependency other than by referencing it from an extra. [This is also by design.](https://github.com/sdispater/poetry/issues/426#issuecomment-421022108)

What is this package needed for again? It's annoying to constantly have an unstaged change in my working copy (I can't install this package because of #352).

trickeydan:
Ok, I have a suggestion. 

As a trial for #97 we could move zoloto to it's own package. This will require a reasonable amount of work, but we need to put it in anyway.

As a temporary solution, I'm ok removing this as a dev package. It's broken on some systems evidently.

Issue #352, opened 2019-07-24
Dependencies packaged as manylinux1 wheels cannot be installed on Nix


kierdavis:
```
[nix-shell:~/checkouts/j5api/j5]$ rm -rf ~/.cache/pypoetry/

[nix-shell:~/checkouts/j5api/j5]$ rm -rf pip-wheel-metadata

[nix-shell:~/checkouts/j5api/j5]$ poetry install
Creating virtualenv j5-py3.7 in /home/kier/.cache/pypoetry/virtualenvs
Installing dependencies from lock file


Package operations: 60 installs, 5 updates, 0 removals

  - Installing zipp (0.5.2)
  - Updating certifi (2018.11.29 -> 2019.6.16)
  - Installing entrypoints (0.3)
  - Installing importlib-metadata (0.18)
  - Installing markupsafe (1.1.1)
  - Installing mccabe (0.6.1)
  - Installing pycodestyle (2.5.0)
  - Installing pyflakes (2.1.1)
  - Updating pyparsing (2.3.1 -> 2.4.0)
  - Installing pytz (2019.1)
  - Updating urllib3 (1.24.1 -> 1.25.3)
  - Installing alabaster (0.7.12)
  - Installing atomicwrites (1.3.0)
  - Updating attrs (18.2.0 -> 19.1.0)
  - Installing babel (2.7.0)
  - Installing docutils (0.15.1)
  - Installing flake8 (3.7.8)
  - Installing imagesize (1.1.0)
  - Installing jinja2 (2.10.1)
  - Installing more-itertools (7.2.0)
  - Installing numpy (1.16.4)
  - Installing pluggy (0.12.0)
  - Installing py (1.8.0)
  - Installing pygments (2.4.2)
  - Updating requests (2.21.0 -> 2.22.0)
  - Installing snowballstemmer (1.9.0)
  - Installing sphinxcontrib-applehelp (1.0.1)
  - Installing sphinxcontrib-devhelp (1.0.1)
  - Installing sphinxcontrib-htmlhelp (1.0.2)
  - Installing sphinxcontrib-jsmath (1.0.1)
  - Installing sphinxcontrib-qthelp (1.0.2)
  - Installing sphinxcontrib-serializinghtml (1.1.3)
  - Installing toml (0.10.0)
  - Installing typing (3.7.4)
  - Installing wcwidth (0.1.7)
  - Installing cached-property (1.5.1)
  - Installing coordinates (0.3.0)
[...]

Issue #97, opened 2019-02-16
Should we have vendor specific implementations in their own packages?


trickeydan:
Should we have competition specific implementations in their own packages? and how far do we go with this?

In my opinion, there should be the following packages for SR as an example:

- `j5` - Core library and component definitions. No specific things should be here.
- `j5.sr` - SR Board and backend definitions. This is where hardware specific things go.
- `sr.robot` - The actual API that competitors interact with. This is where competition specific things go.

kierdavis:
I agree fully with your proposed structure. However I don't think it's worth splitting it up while we're still in such an early stage of development since maintaining code, CI, issues and pull requests across multiple repositories adds a layer of complexity that I feel is unnecessary right now. My suggestion would be to hang on until most things are implemented and we're confident that the project structure isn't going to radically change.

trickeydan:
Ok, I'm marking this as blocked on us writing a lot more code. I realise that this is a somewhat arbitrary thing to block on, please feel free to comment on this issue as necessary.

kierdavis:
I have updated the title to reflect the new direction the issue has taken.

mildlyincompetent:
M o n o r e p o

There is no _technical_ reason that we couldn't have separate python packages in one repo.

That said, I agree with @kierdavis until such time as separate packages become genuinely needed.

trickeydan:
Relevant: #149

trickeydan:
> M o n o r e p o

I'm unsure if you're joking, but a monorepo is probably not a bad idea.

There should still be separate python packages though

kierdavis:
Assuming that we're not intending to transfer maintainership of SR boards to SR's developers (and likewise for other competitions and their specific hardware), I'm onboard for monorepo.

Also note that we should avoid considering boards to be competition-specific. Anyone is free to set up a new competition using whichever boards they choose, and doing so shouldn't incite us to change our package structure. It makes sense to group boards into packages by the principal designer of the board (whether that be hardware/firmware/software), though.

trickeydan:
Following the successful deployment of `j5_zoloto`, I believe consensus has been reached here.

PR #401, opened 2019-09-01
Add CI job for Python 3.8

kierdavis:
While we don't formally support Python 3.8 yet, this will help us
catch any potential incompatibilities ahead of time.

Currently the build fails due to an upstream issue in numpy: https://github.com/numpy/numpy/issues/14403


trickeydan:
This is another case where doing #97 would be useful.

trickeydan:
#402 will probably help with this, as I don't think `j5` is specifically dependent on numpy.

trickeydan:
>  don't think `j5` is specifically dependent on numpy.

We currently depend on PyQuaternion, which depends on numpy.

Thoughts on factoring out vision into a separate module also? It doesn't feel and never has felt like it should be part of the core library.

trickeydan:
The upstream issue has been fixed. I suspect they will release the fix within the next two weeks.

trickeydan:
Numpy has updated.

kierdavis:
I have a feeling that `opencv-contrib-python` isn't available for Python 3.8 and would need to be removed, but it's hard for me to be sure because of #352. Tempted to just see what happens when this hits master and the CI actually gets to run.

trickeydan:
I'm going to merge and see what happens. I'm prepared to remove `opencv-contrib-python` from the dev deps as it shouldn't be there anyway.

#363

trickeydan:
Please see #443

PR #413, opened 2019-09-16
Merge with parent branch instead of master

### LINKED ISSUE #330: Change Merge Check to use parent branch
The tests check if we can merge into master, it should be the parent branch.

trickeydan:
Fixes #330

Uses a `GITHUB_TOKEN` environment variable with an access token from `j5-bot`


.circleci/config.yml
```diff
@@ -13,12 +13,12 @@ jobs:
...
       - checkout
       - run:
-          name: Merge with master
+          name: Merge with parent branch
           command: |
             git config user.name "dummy"
             git config user.email "dummy@example.org"
-            git fetch origin master
-            git merge --squash origin/master
+            git fetch origin $(circleci-branch)
```
kierdavis: I think it would make sense to avoid calling circleci-branch twice if possible:

```bash
parent_branch=$(circleci-branch)
git fetch origin $parent_branch
git merge --squash origin/$parent_branch
```

trickeydan: 8aa6d2d

PR #414, opened 2019-09-16
Fetch the branch we want to merge too

trickeydan:


trickeydan:
I'm seeing the same problem that I saw on #413 here. There seems to be a race condition where the branch is pushed, and then built without CircleCI associating the PR with the build.

Unfortunately, the `circleci-branch` script relies on knowing what the PR is so that it can check against the base branch.

kierdavis:
I believe our circleci setup triggers builds whenever a branch is pushed, which happens before the PR gets created.

It sounds like it's possible to change circleci to instead trigger builds when a PR is created: https://discuss.circleci.com/t/only-build-pull-requests-not-every-branch/200

trickeydan:
I have enabled the appropriate setting in CircleCI and retriggered this build.

PR #415, opened 2019-09-16
Fix broken README code.

### LINKED ISSUE #412: Example Code in README is bad
Since #149, this code uses things which don't exist!

trickeydan:
Fixes #412


README.md
```diff
@@ -31,27 +31,27 @@ The above code is likely to be familiar to any student who has competed in one o
...
+    SRV4MotorBoardHardwareBackend,
+    SRV4PowerBoardHardwareBackend,
+    SRV4RuggeduinoHardwareBackend,
+    SRV4ServoBoardHardwareBackend,
+)
 
 class Robot:
 
     def __init__(self):
+        self.power_board = PowerBoard(SRV4PowerBoardHardwareBackend)
```
kierdavis: `PowerBoard()` should take an *instance* of `SRV4PowerBoardHardwareBackend`, and also a serial number.

```python
power_boards = SRV4PowerBoardHardwareBackend.discover()
if len(power_boards) == 1:
  self.power_board = list(power_boards)[0]
else:
  raise SomeException
```

but at this point you might as well use `BoardGroup.singular`.

README.md
```diff
@@ -31,27 +31,27 @@ The above code is likely to be familiar to any student who has competed in one o
...
+        self.motor_boards = BoardGroup.get_board_group(MotorBoard, SRV4MotorBoardHardwareBackend)
         self.motor_board = self.motor_boards.singular()
 
-        self.servo_boards = BoardGroup(ServoBoard, self._env)
+        self.servo_boards = BoardGroup.get_board_group(ServoBoard, SRV4ServoBoardHardwareBackend)
         self.servo_board = self.servo_boards.singular()
 
-        self.ruggeduino = Ruggeduino(self._env)
-
+        self.ruggeduino = Ruggeduino(SRV4RuggeduinoHardwareBackend)
```
kierdavis: Above comment about `PowerBoard` constructor applies to `Ruggeduino` too.

trickeydan:
> I think we could bring back the `Environment` (in its new form) here, what do you think?

I think we should wait until #417 is merged before continuing.

README.md
trickeydan: a21fe15

README.md
trickeydan: a21fe15

PR #417, opened 2019-09-16
Improve Environments

trickeydan:
Do not merge until #416 is merged

### DIFF
--- a/j5/backends/backend.py
+++ b/j5/backends/backend.py
@@ -173,3 +173,28 @@ class Environment:
             raise NotImplementedError(f"The {str(self)} does not support {str(board)}")
 
         return self.board_backend_mapping[board]
+
+    def merge(self, other: 'Environment') -> None:
+        """
+        Merge in the board-backend mappings from another environment.
+
+        This allows vendors to predefine Environments and API authors
+        can then merge several vendor environments to get the one that
+        they need for their API.
+
+        This method will fail if any board is defined in both environments,
+         as it is unclear which one has the correct mapping.
+        """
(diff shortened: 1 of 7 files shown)


j5/backends/backend.py
```diff
@@ -173,3 +173,28 @@ def get_backend(self, board: Type['Board']) -> Type[Backend]:
...
+        can then merge several vendor environments to get the one that
+        they need for their API.
+
+        This method will fail if any board is defined in both environments,
+         as it is unclear which one has the correct mapping.
+        """
+        boards_present = self.supported_boards | other.supported_boards
+        total_boards = len(self.supported_boards) + len(other.supported_boards)
+
+        if len(boards_present) != total_boards:
```
kierdavis: Would it be simpler to test that the intersection (which you compute below anyway) is non-empty?

tests/backends/test_backend.py
```diff
@@ -121,6 +121,41 @@ def test_environment_board_get_backend_unknown() -> None:
...
+    env2.register_backend(MockBackend)
+    assert len(env2.supported_boards) == 1
+
+    env1.merge(env2)
+    assert len(env1.supported_boards) == 1
+    assert len(env2.supported_boards) == 1
+
+
+def test_environment_merge_duplicate() -> None:
+    """Test that the correct exception is thrown if duplicate entries."""
```
kierdavis:
```suggestion
    """Test that the correct exception is thrown if duplicate entries exist."""
```

j5/backends/backend.py
trickeydan: 0a525ef

PR #418, opened 2019-09-16
Update various bits of the docs

### LINKED ISSUE #392: (Sphinx) Docs Index Not Actually Full Index
The `index` section of the docs actually only indexes the API, not the entire docs. Therefore it should probably live under the API section or be somehow populated to have the entire docs indexed in it.

trickeydan:
Fixes #392 by removing the offending (and useless) page.


kierdavis:
Thanks, this has been needed for a while!

PR #428, opened 2019-10-14
Shorten notation for validating motor powers

tpoomlmly:


j5/components/motor.py
```diff
@@ -61,6 +61,6 @@ def power(self) -> MotorState:
     def power(self, new_power: MotorState) -> None:
         """Set the current state of this output."""
         if isinstance(new_power, float):
-            if new_power < -1 or new_power > 1:
+            if not(-1 < new_power < 1):
```
sedders123: I believe this would error in the case where the power is 1 or -1
```
>>> not(-1 < 1 < 1)
True
>>> not(-1 < -1 < 1)
True
```

compared to
```
>>> 1 < -1 or 1 < 1
False
>>> -1 < -1 or -1 < 1
False
```

I'm also not a huge fan of shortening code for the sake of shortening code. Personally I have to stop and try to parse shorthand boolean logic and it takes me a while to understand intent whereas with the previous syntax its blatantly obvious to a beginner exactly what is being checked.

trickeydan: Quite surprised that this wasn't caught in tests. I should rely on them less when accepting changes.

This should be tested [here](https://github.com/j5api/j5/blob/4e847333a2a93b1c710dad060f2b56553e58a1da/tests/components/test_motor.py#L86) but doesn't appear to for some reason.

trickeydan:
I'm going to pull in master to (hopefully) break the build after #441 merged.

trickeydan:
Looks like the code works now.

I'm going to try and find a style guide that has an opinion on this.

PEP8 doesn't mention it.

sedders123:
While not a style guide I'd argue that these changes go against [The Zen of Python](https://www.python.org/dev/peps/pep-0020/).

Specifically the below points
>Explicit is better than implicit.

>Simple is better than complex.

>Readability counts.

If we find a consistent style guide (or make our own) that we adopt I'm happy to go with whatever that says.

trickeydan:
@tpoomlmly Do you still want to continue with this patch?

tpoomlmly:
I'm good to continue.

sedders123:
Personally, I feel that all we're doing is adding complexity to the project for the sake of saving 6 characters. I think the fact that the first version of this patch would have introduced a [bug](https://github.com/j5api/j5/pull/428#discussion_r334666712) demonstrates the subtleties of potential future issues we could face. If one of our goals is that competitors can understand our code I'd argue the existing code requires less cognitive load to understand. Saying this it wouldn't be the first time I've missed something obvious, @tpoomlmly can you explain why this patch is beneficial?

If this is just a matter of style maybe we should look into a few different python style guides and pick one to adopt?

tpoomlmly:
After sitting and thinking for several hours and consulting with many industry experts and a few hyper-intelligent neural networks, I have come to the conclusion that the original notation was indeed more straightforward. I especially agree with Skybound1's point that the change would only really make sense if the code's checking that new_power's between -1 and 1 rather than outside the range.

PR #429, opened 2019-10-14
Add tooling to extract code snippets from README and typecheck/lint them

### LINKED ISSUE #419: Example code in README can easily fall out of date
I think this has happened more than once. It would we great if we could typecheck it, though I'm not  sure we can do that without pulling it out into a separate file.

kierdavis:
CI will fail pretty badly until #415 gets merged, and even after that it'll still fail because the README code refers to a ruggeduino board that doesn't exist.

Closes #419.


kierdavis:
Closes #419.

trickeydan:
I'm going to update the branch, as it *should* fail with the bad Ruggeduino code in the readme.

trickeydan:
Ok, so it's passing but that's due to #433, which should be fixed before this PR is merged.

kierdavis:
Now that #433 is merged, it's hitting linting errors in the snippet, which I'm currently fixing.

kierdavis:
And now it's hitting the type errors due to missing ruggeduino code, as expected :tada: 

I'm not sure what to do next - I guess either we hold off on merging this until we've implemented the ruggeduino board, or we remove references to the ruggeduino from the example code for the time being. What do you think @trickeydan?

trickeydan:
> And now it's hitting the type errors due to missing ruggeduino code, as expected 
> 
> I'm not sure what to do next - I guess either we hold off on merging this until we've implemented the ruggeduino board, or we remove references to the ruggeduino from the example code for the time being. What do you think @trickeydan?

Remove references in my opinion. It doesn't add much value to the Readme as the servo board example is similar

PR #432, opened 2019-10-15
Set master as the default parent branch

### LINKED ISSUE #430: Master is failing tests
This appears to be due to the `CIRCLE_PULL_REQUEST` variable not being set.

Introduced in #413

trickeydan:
Fixes #430

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,7 +17,8 @@ jobs:
           command: |
             git config user.name "dummy"
             git config user.email "dummy@example.org"
-            parent_branch=$(circleci-branch)
+            branch_output=$(circleci-branch)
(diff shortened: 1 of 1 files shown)


trickeydan:
https://wiki.bash-hackers.org/syntax/pe

kierdavis:
If the pull request env var isn't set, is the merge-with-parent-branch
check even needed?

trickeydan:
> If the pull request env var isn't set, is the merge-with-parent-branch check even needed?

Good point. I think that is a better solution

kierdavis:
> Good point. I think that is a better solution

--> #440

trickeydan:
Closed in favour of #440

PR #443, opened 2019-10-22
Run the CI for Python 3.8

trickeydan:
We only defined it!


trickeydan:
As predicted, 3.8 is failing on `ERROR: Could not find a version that satisfies the requirement opencv-contrib-python==4.1.1.26 (from versions: none)`

I'm going to remove this development dependency. We shouldn't have it in `j5` anyway.

trickeydan:
This PR resolves #363.

PR #445, opened 2019-10-22
Add thread locks to hardware backends

### LINKED ISSUE #381: Check thread-safety
It's not unheard of for teams to use threads in their code to do multiple things at once. We should make sure our robot API is thread-safe.

trickeydan:
This should resolve #381

Locks were chosen over threading.queue or other techniques for their simplicity and ease of implementation.

As mentioned in #381, I have been unable to come up with a good way to unit test this. I am definitely open to ideas, but anything complex should probably go as a separate PR. 

The current recommended test method is integration testing with some weird multi-threaded code.

If we merge this PR without unit tests, a separate issue should be opened to track that.


j5/backends/hardware/sr/v4/power_board.py
```diff
@@ -72,6 +73,8 @@ def discover(cls, find: Callable = usb.core.find) -> Set[Board]:
     def __init__(self, usb_device: usb.core.Device) -> None:
         self._usb_device = usb_device
 
+        self._lock = Lock()
```
kierdavis: Since `RawUSBHardwareBackend` declares and uses `_lock`, could we move this to `RawUSBHardwareBackend.__init__`?

j5/backends/hardware/sr/v4/servo_board.py
```diff
@@ -48,6 +49,7 @@ def discover(cls, find: Callable = usb.core.find) -> Set[Board]:
     @handle_usb_error
     def __init__(self, usb_device: usb.core.Device) -> None:
         self._usb_device = usb_device
+        self._lock = Lock()
```
kierdavis: Since `RawUSBHardwareBackend` declares and uses `_lock`, could we move this to `RawUSBHardwareBackend.__init__`?

j5/backends/hardware/sr/v4/motor_board.py
```diff
@@ -97,6 +100,11 @@ def __del__(self) -> None:
     @handle_serial_error
     def send_command(self, command: int, data: Optional[int] = None) -> None:
         """Send a serial command to the board."""
+        with self._lock:
```
kierdavis: I think the lock only needs to be held for the call to `self._serial.write(bytes(message))`, but it's not a big deal.

j5/backends/hardware/sr/v4/power_board.py
trickeydan: b7bf497

j5/backends/hardware/sr/v4/servo_board.py
trickeydan: b7bf497

j5/backends/hardware/sr/v4/motor_board.py
trickeydan: c09b737

trickeydan:
> It would be ideal if the calls to `set_motor_state` inside `SRV4MotorBoardHardwareBackend.__del__` didn't try to acquire lock, since the code might crash while the lock is held and we don't want `__del__` to deadlock. Maybe we need to add an `acquire_lock` argument that defaults to `True`?

c09b737

j5/backends/hardware/sr/v4/motor_board.py
trickeydan: This was deliberate to try and reduce duplicate code.

If you think that it is important that the lock is only held for the IO operation, I can change it, but context handlers have a side effect of creating duplicate code.

Although we could have a second `null` lock context handler that doesn't do this I suppose..

trickeydan: I'd be interested in some more opinions on this.

@sedders123 @tpoomlmly

trickeydan: I have done 1c6ea21 as an example of my suggestion.

sedders123:
Haven't had a chance to look at this properly (will try to this evening) but my immediate thought is would a `RLock` be more useful here? As `RLock`s allow reentry from the same thread we'd have to worry less about ending up in a deadlock scenario, especially if library authors/competitors do something weird.

I haven't had time to fully think this through so there could be something I'm missing, but thought I'd share to get others opinions.

trickeydan:
> Haven't had a chance to look at this properly (will try to this evening) but my immediate thought is would a `RLock` be more useful here? As `RLock`s allow reentry from the same thread we'd have to worry less about ending up in a deadlock scenario, especially if library authors/competitors do something weird.
> 
> I haven't had time to fully think this through so there could be something I'm missing, but thought I'd share to get others opinions.

I'll have a look at this later, you might be right. Worth noting that sr.robot uses Lock https://github.com/srobo/sr-robot/blob/4e6eebf23cb7e3f70f123f92d15b6c5cc772b17d/sr/robot/motor.py#L49

Robotd did not implement locks.

j5/backends/hardware/sr/v4/motor_board.py
kierdavis: Hmm, so looking at the current code again now it's obvious that's how `_send_command_no_lock` is being used (I was aware of this when I originally reviewed the code too so I have no idea what was going through my head when I left that comment).

1c6ea212b37fbf23b51c1eea0daf267452e945cc is a neat solution, but I think it's less straightforward than what we had originally - I think the original solution with `_send_command_no_lock` is easier to understand.

Sorry for the wild goose chase here.

trickeydan: 9ce14b9

PR #446, opened 2019-10-22
Use good serial driver for graceful shutdown in Bad Write tests.

### LINKED ISSUE #394: Investigate strange exception in MB tests
```
tests/backends/hardware/sr/v4/test_motor_board.py .......Exception ignored in: <function SRV4MotorBoardHardwareBackend.__del__ at 0x7fa76805a8c0>
Traceback (most recent call last):
  File "/home/dan/Code/sro/j5/j5/backends/hardware/sr/v4/motor_board.py", line 94, in __del__
    self.set_motor_state(i, MotorSpecialState.BRAKE)
  File "/home/dan/Code/sro/j5/j5/backends/backend.py", line 32, in new_method
    retval = old_method(*args, **kwargs)
  File "/home/dan/Code/sro/j5/j5/backends/hardware/sr/v4/motor_board.py", line 153, in set_motor_state
    self.send_command(CMD_MOTOR[identifier], value)
  File "/home/dan/Code/sro/j5/j5/backends/hardware/j5/serial.py", line 34, in catch_exceptions
    return func(*args, **kwargs)
  File "/home/dan/Code/sro/j5/j5/backends/hardware/sr/v4/motor_board.py", line 108, in send_command
    "Mismatch in command bytes written to serial interface.",
j5.backends.backend.CommunicationError: Mismatch in command bytes written to serial interface.
```

trickeydan:
Fixes #394

The bad serial driver was being used to destruct the Motor Driver which
was causing an obscure issue.

This took a while to debug. The flow of code was not obvious. I just
looked at the constructor and assumed that it was using MotorSerial
throughout


kierdavis:
This looks like a good use case for https://docs.python.org/3/library/unittest.mock.html#unittest.mock.patch

trickeydan:
> This looks like a good use case for [docs.python.org/3/library/unittest.mock.html#unittest.mock.patch](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.patch)

I've been unable to get this to work in this instance.

PR #471, opened 2019-11-26
Refactor to remove handle_usb_decorator

### LINKED ISSUE #449: Replace handle_usb_error decorator
Related: #395, #448 

Decorators break typing.

trickeydan:
Fixes #449

### DIFF
--- a/j5/backends/backend.py
+++ b/j5/backends/backend.py
@@ -68,7 +68,7 @@ class BackendMeta(ABCMeta):
             return cls
 
         # Check if this is an abstract Backend.
-        if len(cls.__bases__) == 1 and cls.__base__.__name__ != "Backend":
+        if hasattr(cls, "ABSTRACT") and cls.ABSTRACT:
             return cls
 
         mcs._check_component_interfaces(cls)  # type: ignore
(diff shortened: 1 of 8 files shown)


j5/backends/backend.py
```diff
@@ -68,7 +68,7 @@ def __new__(mcs, name, bases, namespace, **kwargs):  # type:ignore
             return cls
 
         # Check if this is an abstract Backend.
-        if len(cls.__bases__) == 1 and cls.__base__.__name__ != "Backend":
+        if hasattr(cls, "ABSTRACT") and cls.ABSTRACT:
```
kierdavis:
```
>>> class Abstract(metaclass=ABCMeta):
...     @abstractmethod
...     def my_method(self):
...             pass
... 
>>> class NotAbstract(Abstract):
...     def my_method(self):
...             pass
... 
>>> bool(Abstract.__abstractmethods__)
True
>>> bool(NotAbstract.__abstractmethods__)
False
```

trickeydan: d9c79d4

Have tested and it appears to work! :+1:

PR #493, opened 2020-01-10
Update author email

trickeydan:


kierdavis:
Oops, didn't mean to hit merge.

Does anyone else object to this?

trickeydan:
It'll hit the mailing list, rather than SRO. I don't think anybody on the j5 team has access to the SRO email anymore, so we need to change it to *something*.

Issue #489, opened 2020-01-02
Multiple layers of inheritance break bound TypeVars


trickeydan:
```
U = TypeVar('U', bound=Backend)

def get_board_group(cls, _: Type[T], backend: Type[U]) -> 'BoardGroup[T, U]':
   ...
```

If a non-direct subclass of `Backend` is passed in as `U`, it will break.

```
error: Value of type variable "U" of "get_board_group" of "BoardGroup" cannot be "SRV4MotorBoardHardwareBackend"
```

This is possibly an issue in `mypy` that was introduced between version `0.740` and `0.750`. We should investigate and report upstream.

PR #513, opened 2020-01-27
Remove the serial handler decorator

### LINKED ISSUE #448: Replace handle_serial_error
Decorators break type checking.

Let's do this properly!

trickeydan:
Fixes #448


trickeydan:
Many of the `type: ignore` statements that are added in this PR are due to #489

trickeydan:
> I wonder if a context manager would do a better job of reducing the duplicated try..except blocks without breaking typing like decorators do...

Context Managers would definitely circumvent the issues with typing that decorators suffer from. I think we should investigate this is the future.

PR #520, opened 2020-01-28
Add tests for the Orientation class

### LINKED ISSUE #480: No test coverage on Orientation class
This class is not tested.

Related: #468

### LINKED ISSUE #519: rot_x, rot_y & rot_z on Orientation are the wrong way round
The X axis is side-to-side, so rot_x should be pitch 
Y should be yaw 
Z should be roll

tpoomlmly:
Fixes #480 and Fixes #519


tests/vision/test_orientation.py
```diff
@@ -0,0 +1,70 @@
+"""Tests for orientation class."""
+
+from numpy import allclose
+from pyquaternion import Quaternion
+
+from j5.vision.orientation import Orientation
+
+
+def test_quaternion() -> None:
```
trickeydan: Could you update these function names to match the convention we have in the rest of the tests?

For example: `def test_coordinate_from_cylindrical() -> None:`

Therefore, this one would be:

```suggestion
def test_orientation_quaternion() -> None:

```

PR #521, opened 2020-01-29
Add full stops and capitals to exception messages in power board.

### LINKED ISSUE #496: Typo in Power Board Exception
https://github.com/j5api/j5/blob/4347e5bb3e573748f3b7387a640bcb382226af68/j5/backends/hardware/sr/v4/power_board.py#L127

trickeydan:
Fixes #496

### DIFF
--- a/j5/backends/hardware/sr/v4/power_board.py
+++ b/j5/backends/hardware/sr/v4/power_board.py
@@ -105,8 +105,10 @@ class SRV4PowerBoardHardwareBackend(
         try:
             return self._output_states[identifier]
         except KeyError:
-            raise ValueError(f"Invalid power output identifier {identifier!r}; "
-                             f"valid identifiers are {CMD_WRITE_OUTPUT.keys()}") from None
+            raise ValueError(
+                f"Invalid power output identifier {identifier!r}; "
+                f"valid identifiers are {CMD_WRITE_OUTPUT.keys()}.",
+            ) from None
 
     def set_power_output_enabled(
         self, identifier: int, enabled: bool,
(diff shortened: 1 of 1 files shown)


j5/backends/hardware/sr/v4/power_board.py
```diff
@@ -116,7 +116,7 @@ def set_power_output_enabled(
             cmd = CMD_WRITE_OUTPUT[identifier]
         except KeyError:
             raise ValueError(f"Invalid power output identifier {identifier!r}; "
```
tpoomlmly: Line 121 is the culprit that's failing the tests. I suggest moving the string onto a new line, although that will reduce consistency.
```suggestion
            raise ValueError(
                f"Invalid power output identifier {identifier!r}; "
                f"valid identifiers are {CMD_WRITE_OUTPUT.keys()}."
            ) from None
```

tpoomlmly: There's also no full stop on line 109, but adding one will push it over 90 characters long.

tpoomlmly: Line 119 is the culprit that's failing the tests. I propose putting the string on a new line, although that will reduce consistency.
```suggestion
            raise ValueError(
                f"Invalid power output identifier {identifier!r}; "
                f"valid identifiers are {CMD_WRITE_OUTPUT.keys()}."
            ) from None
```

tpoomlmly: There's also a missing full stop on line 109, but adding it will push the line over 90 characters.

trickeydan: > although that will reduce consistency.

We don't mind as long as it passes the linter :)

PR #523, opened 2020-01-29
Improve Servo Board Backend Tests

### LINKED ISSUE #312: Servo board tests are insufficient
Given that #308 happened, we need to make the tests on the servo board better and ensure that the tests for the other boards are sufficient.

trickeydan:
Fixes #312

There are a couple of `type: ignore`s in the tests. This is due to deliberately passing in a broken type, as students may not run a type checker.

### DIFF
--- a/j5/backends/hardware/sr/v4/servo_board.py
+++ b/j5/backends/hardware/sr/v4/servo_board.py
@@ -62,8 +62,7 @@ class SRV4ServoBoardHardwareBackend(
         ]
 
         # Initialise servos.
-        with self._lock:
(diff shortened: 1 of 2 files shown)


trickeydan:
Codefactor is raising an issue for no good reason. I'll change it to pass anyway *sigh*

trickeydan:
Ok, so Codefactor doesn't seem to be syncing with Github properly. :(

trickeydan:
> Ok, so Codefactor doesn't seem to be syncing with Github properly. :(

I've overidden it for this PR.

Issue #522, opened 2020-01-29
`discover` method should inherit to change things


trickeydan:
Applies to at least the Power and Servo board.

This is not pythonic:

`def discover(cls, find: Callable = usb.core.find) -> Set[Board]:`

We should be subclassing it in tests. This will also help fix typing issues.

Also the same in ConsoleBackends

Related to #404

trickeydan:
This is going to be hard. Just had a go and failed.

PR #528, opened 2020-02-01
Add extra flake8 plugins to linting

### LINKED ISSUE #499: Additional flake8 plugins
Thoughts on adding:

- `flake8-bugbear`
- `flake8-builtins`
- Possibly `flake8-print`

trickeydan:
- `flake8-print`
- `flake8-builtins`
- `flake8-bugbear`

Modifications other than adding these are explained in commit messages.

Fixes #499

### DIFF
--- a/j5/backends/console/console.py
+++ b/j5/backends/console/console.py
@@ -8,11 +8,14 @@ T = TypeVar("T")
 class Console:
     """A helper class for console backends."""
 
+    DEFAULT_PRINT: Callable = print  # type: ignore  # noqa: T002
+    DEFAULT_INPUT: Callable = input  # type: ignore
+
     def __init__(
             self,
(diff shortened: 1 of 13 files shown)


j5/backends/console/console.py
```diff
@@ -8,11 +8,14 @@
 class Console:
     """A helper class for console backends."""
 
+    DEFAULT_PRINT: Callable = print  # type: ignore  # noqa: T002
```
sedders123: Nit: Should these class variables be preceded by an underscore to indicate we don't expect anyone to use them outside of maybe tests?

trickeydan: Probably, yeah.

I'm planning a refactor of this class anyway, similar to #522 as it could definitely be more pythonic.

PR #530, opened 2020-02-03
Use cached_property in Coordinate class

### LINKED ISSUE #350: Consider use of cached_property
https://pypi.org/project/cached-property/

trickeydan:
Fixes #350


trickeydan:
This will probably magically fix itself when #528 is merged and rebased.

Issue #549, opened 2020-03-02
Use Tuple[Component] on boards to prevent mutability issues.


tpoomlmly:
It's possible to do this:
`Robot().motor_board.motors[0] = 1` rendering the motor unusable.

trickeydan:
This is a wider problem that stems from using `List[Motor]`, as `List` is mutable and therefore won't stop them from editing it.

As this is currently the syntax of the SR API, it is very important that we change this.

However, this is a breaking change due to the type change, therefore rescheduling it to `0.10.x`

PR #572, opened 2020-05-29
Fix mutability issues on components

trickeydan:
Also test for this on BoardGroups too.

This PR changes some lists to tuples and some dicts to an immutable dict type.

The primary reason for this is for when the following student mistake is made:

```python
r.motor_boards[0].motors[0] = 100
```

Correct code:

```python
r.motor_boards[0].motors[0].power = 100
```

This is likely to occur a lot as it's similar to the old SR API.

### DIFF
--- a/tests/boards/sr/v4/test_power_board.py
+++ b/tests/boards/sr/v4/test_power_board.py
@@ -2,6 +2,8 @@
 from datetime import timedelta
 from typing import TYPE_CHECKING, Optional, Set
 
+import pytest
+
 from j5.backends import Backend
 from j5.boards.sr.v4 import PowerBoard, PowerOutputGroup, PowerOutputPosition
 from j5.components import (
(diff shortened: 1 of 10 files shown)


trickeydan:
Fixes #549

tests/boards/sr/v4/test_power_board.py
```diff
@@ -181,3 +183,15 @@ def test_power_board_wait_start() -> None:
     pb.wait_for_start_flash()
 
     assert pb._run_led.state
+
+
+def test_output_mutability() -> None:
+    """
+    Test the mutability of Motors.
```
tpoomlmly, line 190: The docstrings of `test_output_mutability()` and `test_servo_mutability()` both refer to motors instead of outputs and servos, respectively.

trickeydan: de91bd3

PR #595, opened 2020-07-05
Add Backend as superclass of SerialHardwareBackend

trickeydan:
Spotted by @sersorrel in https://github.com/srobo/sr-robot3/pull/2#discussion_r449811261

This also means that the abstract methods are now actually enforced.

We should consider introducing a run-time type check for this, but I am
unable to think of a good place to put it at the moment.

## Checklist

Stops need to cast to `Type[Backend]` in environment definitions.

`SerialHardwareBackend` and subclasses thereof, `SRV4MotorBoardBackend`, `SBArduinoHardwareBackend`

Shouldn't make any difference. This is purely for typing

We had previously attributed this bug to #489, although didn't have an open issue for it :/

### DIFF
--- a/j5/backends/hardware/j5/serial.py
+++ b/j5/backends/hardware/j5/serial.py
@@ -6,7 +6,7 @@ from typing import Optional, Set, Type
 from serial import Serial, SerialException, SerialTimeoutException
 from typing_extensions import Protocol
 
-from j5.backends import BackendMeta, CommunicationError
+from j5.backends import Backend, BackendMeta, CommunicationError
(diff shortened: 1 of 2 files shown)


sersorrel:
mypy's now unhappy because there's a `# type: ignore` which isn't needed any more, added [here](https://github.com/j5api/j5/commit/d892f6a5b89ab9e4271fe257cd0f9547aef5197b#diff-a84e50d64ab15a245c140ac9ec159c77L74-R74) as part of #513, fwiw.

PR #600, opened 2020-07-06
Use explicit methods for controlling GPIOPin

### LINKED ISSUE #592: Use methods on GPIOPin for more explicit digital/analogue_read/write
e.g `pin.digital_read()`

trickeydan:
Adds following methods:

- `digital_write`
- `digital_read`
- `analogue_write`
- `analogue_read`

Also `last_digital_write` property.

These replace the previous behaviours of `digital_state` and
`analogue_value`.

## Checklist

Previous ways to control the `GPIOPin` component were designed to be very consistent with other parts of the API, for example on the `Motor` component, where property setters and getters are used heavily.

This resulted in confusing functionality, where the same property would behave differently depending on the `mode` of the `GPIOPin`, which will likely result in a high rate of student bugs. For example:

```python

pin = r.ruggeduino.pins[3]

pin.mode = DIGITAL_WRITE

print(pin.digital_state)
```

```python

pin = r.ruggeduino.pins[3]

pin.mode = DIGITAL_READ

print(pin.digital_state)
```

Both of these snippets are very similar, both will print either `True` or `False`. However, the first just reads the cached last set state, and the latter actually performs a digital read.

- `GPIOPin`, and its tests.
- `tests_hw/sb_arduino`

Previous functionality was confusing, where the same property would behave differently depending on the `mode` of the `GPIOPin`, which will likely result in a high rate of student bugs. 

Closes #592

These changes were also discussed by the SR Brain Team in a meeting on 2nd July, where it was concluded that the API should change. This new API is more similar to `sr-robot` and `robot-api` too.

### DIFF
--- a/j5/components/gpio_pin.py
+++ b/j5/components/gpio_pin.py
@@ -187,22 +187,27 @@ class GPIOPin(Component):
         """Set the analogue value of the pin."""
         self._require_pin_modes({
             GPIOPinMode.ANALOGUE_OUTPUT,
-            GPIOPinMode.PWM_OUTPUT,
         })
         if new_value < 0 or new_value > 1:
             raise ValueError("An analogue pin value must be between 0 and 1.")
 
-        if self.mode is GPIOPinMode.ANALOGUE_OUTPUT:
-            self._backend.write_gpio_pin_dac_value(
-                self._identifier,
-                new_value,
-            )
-        else:
-            # We must be a PWM_OUTPUT
-            self._backend.write_gpio_pin_pwm_value(
-                self._identifier,
-                new_value,
-            )
+        self._backend.write_gpio_pin_dac_value(
+            self._identifier,
+            new_value,
+        )
+
+    def pwm_write(self, new_value: float) -> None:
+        """Set the PWM value of the pin."""
+        self._require_pin_modes({
+            GPIOPinMode.PWM_OUTPUT,
+        })
+        if new_value < 0 or new_value > 1:
+            raise ValueError("An PWM pin value must be between 0 and 1.")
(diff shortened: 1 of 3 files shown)


j5/components/gpio_pin.py
```diff
@@ -152,36 +152,38 @@ def mode(self, pin_mode: PinMode) -> None:
...
-    @property
-    def analogue_value(self) -> float:
+    def analogue_read(self) -> float:
         """Get the scaled analogue reading of the pin."""
         self._require_pin_modes({GPIOPinMode.ANALOGUE_INPUT})
         return self._backend.read_gpio_pin_analogue_value(self._identifier)
 
-    @analogue_value.setter
-    def analogue_value(self, new_value: float) -> None:
+    def analogue_write(self, new_value: float) -> None:
```
kierdavis, line 186: I feel like it would be in the spirit of this PR to move the pwm functionality here into a separate `pwm_write` method - what do you think?

tpoomlmly: It could also be useful to have `last_analogue_write` and `last_pwm_write` methods to go with `last_digital_write`, though these would need modifications to the backends and might be a bit too out of scope.

trickeydan: b6b5cf6

trickeydan: > It could also be useful to have `last_analogue_write` and `last_pwm_write` methods to go with `last_digital_write`, though these would need modifications to the backends and might be a bit too out of scope.

I agree with this idea, it does require changes to the backend though, so will open a separate issue.

Issue #250, opened 2019-04-18
Should "safe state" of motor be BRAKE or COAST?


kierdavis:
Spun out of #227; see [discussion](https://github.com/j5api/j5/pull/227#discussion_r275446457) there.

A third option is to leave this decision down to the competition vendor to decide.

mildlyincompetent:
Generally speaking, when someone is stabbed, it is best practice *not* to remove the stabbing instrument prior to medical personnel being present, the reason for this being simply that the removal adds further movement to a stab wound, which generally causes bad things to occur.

Similarly, if one is working with a giant, deadly, spinny sawblade of death, you would much rather it come to an abrupt stop if something goes wrong as opposed to coasting its way through your arm and/or face.

Now of course, the robots we use do not generally carry stabbing paraphernalia or spinny sawblades of death, but the general point I feel stands - if your hand is trapped in a mechanism, that mechanism ceasing to move immediately guarantees no further damage is done by the mechanism to your hand. Sure, your hand may still be trapped, but that's better than it being, say, detached. It is then trivial to transition to a state of coast simply by cutting power to the motors. Consider the alternative - the motor coasts and may result in further damage.

To me it seems like BRAKE guarantees that the robot cannot do further damage, while potentially being a bit annoying to those damaged. COAST on the other hand may cause further damage, which is the primary thing we would like to avoid, hence I suggest using BRAKE as the safe state.

kierdavis:
Come to think of it, this choice doesn't actually matter for our current use cases. The power board will cut power to all of its outputs including any motor boards, so all motors will coast regardless of what you tell the motor board to do.

If we still want to make a (theoretical) decision on this, I think we need to come up with some realistic scenarios. As exciting as spinny sawblades of death or stabbing paraphernalia sound they're not good examples of the typical (or even worst-case) safety issues that come about during our current activities.

trickeydan:
I have come up with the following scenarios for us to consider. Feel free to add more.

## Scenario A

A robot with two tank tracks and 4 motors is travelling at maximum speed during tinker time. Due to the volunteers being busy, there are a large number of students inside the arena at this time. The robot has not been safety checked due to human error, and may contain sharp implements. The LiPo battery is close to the motor, and not particularly secure.

[...]

PR #606, opened 2020-07-18
Allow specification of the safe state of the Motor Board

### LINKED ISSUE #304: Allow safe_state of motor to be set by API Author
Spun out of #250.

Default should probably be brake, but overridable.

trickeydan:
Fixes #304

## Checklist

Discussed in #250 and #304 

`SRV4MotorBoard`

It shouldn't at all. Only affected API Authors in a backwards-compatible manner.

Fixes #304

### DIFF
--- a/j5/boards/sr/v4/motor_board.py
+++ b/j5/boards/sr/v4/motor_board.py
@@ -3,7 +3,12 @@ from typing import TYPE_CHECKING, Optional, Set, Type, cast
 
 from j5.backends import Backend
 from j5.boards import Board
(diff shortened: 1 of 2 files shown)


j5/components/motor.py
```diff
@@ -38,6 +38,7 @@ def __init__(
             self,
             identifier: int,
             backend: MotorInterface,
+
```
tpoomlmly, line 41: What's the reasoning behind the gap?

trickeydan: None. Force pushed.

PR #616, opened 2020-07-28
Rename Board.serial to Board.serial_number

### LINKED ISSUE #593: Rename serial to serial_number
Can possibly tie in with stuff from #368 

That would make serial numbers an interface, rather than an intrinsically required thing.

trickeydan:
## Checklist

- Makes the difference between a UART serial and serial number clear.
- Requested change from SR @j5api/brain-team 

Anything that uses a serial number

Makes the difference between a UART serial and serial number clear.

Fixes #593

### DIFF
--- a/docs/_code/qs_adding_boards.py
+++ b/docs/_code/qs_adding_boards.py
@@ -24,7 +24,7 @@ class MyRobot(BaseRobot):
 
 r = MyRobot()
 
-print(f"Found Power Board: {r.power_board.serial}")
+print(f"Found Power Board: {r.power_board.serial_number}")
 print(f"Power Board Firmware: {r.power_board.firmware_version}")
 
 # Access a board specific function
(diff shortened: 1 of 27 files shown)


j5/boards/sb/arduino.py
```diff
@@ -24,6 +25,30 @@ def __init__(
...
+        pins = ImmutableDict[ArduinoUno.PinNumber, GPIOPin]({
+            **cast(Mapping[ArduinoUno.PinNumber, GPIOPin], self._analogue_pins),
+            **cast(Mapping[ArduinoUno.PinNumber, GPIOPin], self._digital_pins),
+
+        })
+        return pins
+
+    def make_safe(self) -> None:
+        """Make this board safe."""
+        pass
```
sersorrel: this looks like an unrelated change?

trickeydan: Literally force-pushed just now to get rid of that 😅

trickeydan: (Was an artifact of a bodged rebase)

Issue #715, opened 2021-06-19
Should we de-scope vision?


trickeydan:
I'd like to propose that we de-scope vision from `j5` and remove all related code. We can point developers in the direction of Zoloto and other similar libraries.

**Pros**
- We are currently duplicating half of the human friendly API from Zoloto and this is a pain to maintain.
- Less effort to implement vision on APIs, especially as it's usually a custom use case
- We only support one vision system anyway, and it's currently broken!
- We can remove a lot of untested code
- No dependencies on OpenCV or pyquaternion, or numpy

**Cons**
- Less features

RealOrangeOne:
Disclaimer: As the maintainer of [Zoloto](https://github.com/RealOrangeOne/zoloto), my opinion may be bias, but equally educated.

I like the idea of removing the duplication of code, there's no need to have the duplication and wrapper just for the sake of a generic and composable API. It also means we can focus developer time into creating features and improving the platform, rather than creating shims for things which already exist.

With that said, there is a way around this which may be better: A strict dependency on zoloto (optional). Therefore just working under the assumption that "j5 vision uses zoloto and only zoloto". Therefore rather than specific j5 APIs, it's just exposing Zoloto constructs, with perhaps some wrappers to support j5s different backends (these would likely be very simple). Obviously there are lots of cases where vision isn't required, and OpenCV can be a big dependency, so probably want it as an optional dependency (`j5[vision]`).

Zoloto has specifically been designed to be generic, widely useful, and highly configurable, so it should fit nicely with the feature and composability requirements and ethos of j5.

sedders123:
My gut says descope it, but gonna spew some thoughts below and see where I end up.

<hr>

> * We are currently duplicating half of the human friendly API from Zoloto and this is a pain to maintain.

Our own Vision API does allow us to shim around Zoloto breaking changes and give us the ability to switch vision provider if we wanted to.
> * Less effort to implement vision on APIs, especially as it's usually a custom use case

I think this means that consumers of `j5` would have more fine grained control over how there visions system works as they would implement it themselves. The flipside of that is that if consumers want a vision system they have to implement their own.
> * We only support one vision system anyway, and it's currently broken!

We only support one power board at the moment. We could fix our current vision system.
> * We can remove a lot of untested code

Or we could test it
> * No dependencies on OpenCV or pyquaternion, or numpy

Fewer dependencies would be nice, but if we're expecting consumers to roll their own vision they'll likely end up with these anyway.

<hr>

[...]

PR #718, opened 2021-07-03
Remove vision support

### LINKED ISSUE #346: Add tests for Zoloto

### LINKED ISSUE #402: Separate Zoloto into it's own package
Ok, I have a suggestion. 

As a trial for #97 we could move zoloto to it's own package. This will require a reasonable amount of work, but we need to put it in anyway.

As a temporary solution, I'm ok removing this as a dev package. It's broken on some systems evidently.

_Originally posted by @trickeydan in https://github.com/j5api/j5/issues/363#issuecomment-526943265_

trickeydan:
## Checklist

Following discussions in #715, I believe the best way forward is to move vision code into a `j5-zoloto` module.

I intend to directly copy all of the removed code to a separate repo, and will attempt to preserve history if possible.

De-coupling zoloto and j5 will allow us to resolve issues with duplicated data structures and suspected issues converting between them.

By removing untested code, this PR increases test coverage to 97%. We will be able to test the zoloto integration 

Remove all vision code. Remove dependency on `pyquaternion` and `cached-property`

Vision is a complex problem and is often very specific to a particular competition. I would be interested in re-integrating parts of common vision code in the future, but only once we know it works and without adding a dependency on Zoloto.

This change will allow us to focus on the current codebase and move towards `1.0`. Vision is best left to the experts for now.

Closes #402 and closes #346


RealOrangeOne (approved):
Overall I think this looks fine (with my experience of the project)

That said, it looks like there are some pretty useful looking tests for some of the constructs. Are these definitely also covered upstream? It'd be great to not lose that work!

trickeydan:
> That said, it looks like there are some pretty useful looking tests for some of the constructs. Are these definitely also covered upstream? It'd be great to not lose that work!

We definitely don't want to lose any of this code, we're just moving it elsewhere :) https://github.com/j5api/j5-zoloto now exists so most of it will go there. I'm trying to work out how we can preserve the author history of the code removed in the new repo this morning

sedders123 (approved):

raccube (approved):

PR #751, opened 2021-11-09
Drop numpy as a dev dependency

trickeydan:
## Checklist

This is now a dependency in j5-zoloto, so it is not needed here.

Also, we were locking it to a sufficiently old version of numpy that it wouldn't install the dependencies on darwin-arm64.

None

M1 Macs can now be used for development

Nope.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -40,7 +40,6 @@ isort = {version = "^5.6",extras = ["pyproject"]}
 flake8-bugbear = "^21.9.2"
 flake8-builtins = "^1.5.3"
 flake8-print = "^4.0.0"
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -17,7 +17,7 @@ classifiers = [
 ]
 
 [tool.poetry.dependencies]
-python = "^3.6"
+python = "^3.7"
```
RealOrangeOne, line 20: Question: Is this change intentional?

trickeydan: Fixed and force-pushed :P

RealOrangeOne (approved):

raccube (approved):

PR #755, opened 2021-11-17
Add an additional ruggeduino USB id

RealOrangeOne:

So these ruggeduinos can be detected

> Bus 003 Device 048: ID 16d0:0613 MCS Ruggeduino

The files that I changed

It means their ruggeduinos can be detected


trickeydan (approved):
Not sure why the build is failing, will investigate later

Issue #758, opened 2021-11-28
Ensure arduino pins are set as outputs when changing mode


RealOrangeOne:
Currently, when setting a pin as `OUTPUT`, the arduino never gets the message. For all other pins it works as expected.

The cause of this is in:

https://github.com/j5api/j5/blob/4517772684615516e73a09edad751df6ed17bb44/j5/backends/hardware/sr/v4/ruggeduino.py#L103-L107

Note how only the state is set, not the pin mode, which causes weird things to happen. This could be patched with firmware, but that's a lot of work, and arguably fixing it here is the correct course of action anyway.

WillB97:
There are kind of 2 issues here:
1. `_update_digital_pin` doesn't set the output mode
2. Setting the pin mode and setting the output state should be separate functions as we don't want to repeatedly set the output mode

trickeydan:
Issue confirmed.

PR #762, opened 2021-12-02
Fix Ruggeduino driver not setting DIGITAL_OUTPUT mode

trickeydan:
## Checklist

Teams couldn't set their pins. See #758 

Ruggeduino hardware backend, and associated test suite.

Yes.

#758


trickeydan:
Associated FW: https://github.com/srobo/ruggeduino-fw/blob/master/src/ruggeduino.cpp

WillB97:
While this looks like it likely works, I think that there is still an implementation issue here. Fundamentally the pin's mode and value are separate and distinct property and we should be settng them independently.

trickeydan:
> While this looks like it likely works, I think that there is still an implementation issue here. Fundamentally the pin's mode and value are separate and distinct property and we should be settng them independently.

That's something we can fix, but is a separate issue with how the API currently works.

I have ideas for how we can fix it, but it's out of scope for this PR in my opinion. It's probably better to make students' code work  first.

Please see the commit message for a detailed explanation of why.

WillB97 (approved):

WillB97 (approved):

RealOrangeOne (approved):

PR #829, opened 2022-10-14
Handle ruggeduino occasionally returning spurious bytes

WillB97:
## Checklist

- [ ] I have updated the relevant documentation
- [ ] I have added the `semver-major`, `semver-minor` or `semver-patch` label as appropriate

The ruggeduino occasionally returns "\xFF" when first queried which can't be converted to unicode, so throws a UnicodeDecodeError during discovery.

Serial devices.

### DIFF
--- a/j5/backends/hardware/j5/serial.py
+++ b/j5/backends/hardware/j5/serial.py
@@ -1,4 +1,5 @@
 """Abstract hardware backend implementation provided by j5 for serial comms."""
+import logging
 from abc import abstractmethod
 from datetime import timedelta
 from typing import List, Optional, Set, Type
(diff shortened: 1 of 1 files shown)


j5/backends/hardware/j5/serial.py
```diff
@@ -92,7 +93,14 @@ def read_serial_line(self, empty: bool = False) -> str:
...
 
-        ldata = bdata.decode('utf-8')
+        try:
+            ldata = bdata.decode('utf-8')
+        except UnicodeDecodeError as e:
+            if empty:
+                logging.getLogger(__file__).error(f"{e} in {bdata!r}")
+                ldata = ''
+            else:
+                raise
```
sedders123, line 104: Nit: Think we can short circuit here
```suggestion
                return ''
            raise
```

sedders123 (approved):

